PR #183, opened 2024-06-11
Combine question ded

zstarkuw:
Changed a column header and renamed some files for consistency.
Added script to concatenate form CSVs.
Committed initial run of DED script.


kgauthreaux (commented):
Looks good! Is it possible for the header form variables to be at the top?

zstarkuw:
> Looks good! Is it possible for the header form variables to be at the top?

updated

kgauthreaux (approved):

PR #189, opened 2024-08-23
A2 header variables checks

kgauthreaux:
Adding header variable error checks to A2


forms/uds/a2/form_a2_ivp_error_checks_mc.csv
```diff
@@ -1,26 +1,40 @@
...
+a2-ivp-c-029,029,Error,a2,I,LANGA2,Conformity,LANGA2 conforms,LANGA2 must be an integer between 1-2,Q0c. LANGA2 (Language of A2 administration) must be an integer between 1-2,If LANGA2 < 1 or LANGA2 > 2,,,Yes,Yes,
+a2-ivp-m-030,030,Error,a2,I,MODEA2,Missingness,MODEA2 must be present,MODEA2 cannot be blank,Q0d. MODEA2 (A2 in person or remote) cannot be blank,If MODEA2 = blank,,,Yes,No,
+a2-ivp-c-031,031,Error,a2,I,MODEA2,Conformity,MODEA2 conforms,MODEA2 must be an integer between 1-3,Q0d. MODEA2 (A2 in person or remote or not complete) must be an integer between 1-3,If MODEA2 < 1 or MODEA2 > 3,,,Yes,No,
+a2-ivp-m-032,032,Error,a2,I,RMREASA2,Missingness,RMREASA2 conditionally present,"If MODEA2 = 2, RMREASA2 cannot be blank",If Q0d. MODEA2 (A2 in person or remote) = 2 (Remote) then Q0d1. RMREASA2 (Reason for remote A2) cannot be blank,If MODEA2 = 2 and RMREASA2 = blank,,,Yes,No,
+a2-ivp-m-033,033,Error,a2,I,RMREASA2,Missingness,RMREASA2 conditionally blank,"If MODEA2 = 1, RMREASA2 must be blank",If Q0d. MODEA2 (A2 in person or remote) = 1 (In-person) then Q0d1. RMREASA2 (Reason for remote A2) must be blank,If MODEA2 = 1 and RMREASA2 not = blank,,,Yes,No,
```
hewanadun, line 34: Shouldn't this check be "If MODEA2 = 1 or 3" ?

forms/uds/a2/form_a2_ivp_error_checks_mc.csv
```diff
@@ -1,26 +1,40 @@
...
+a2-ivp-c-034,034,Error,a2,I,RMREASA2,Conformity,RMREASA2 conforms,RMREASA2 must be an integer between 1-5,Q0d1. RMREASA2 (Reason for remote A2) must be an integer between 1 and 5,If RMREASA2 < 1 or RMREASA2 > 5,,,Yes,No,
+a2-ivp-m-035,035,Error,a2,I,RMMODEA2,Missingness,RMMODEA2 conditionally present,"If MODEA2 = 2, RMMODEA2 cannot be blank",If Q0d. MODEA2 (A2 in person or remote) = 2 (Remote) then Q0d1a. RMMODEA2 (Mode of remote A2 administration) cannot be blank,If MODEA2 = 2 and RMMODEA2 = blank,,,Yes,No,
+a2-ivp-m-036,036,Error,a2,I,RMMODEA2,Missingness,RMMODEA2 conditionally blank,"If MODEA2 = 1, RMMODEA2 must be blank",If Q0d. MODEA2 (A2 in person or remote) = 1 (In-person) then Q0d1a. RMMODEA2 (Mode of remote A2 administration) must be blank,If MODEA2 = 1 and RMMODEA2 not = blank,,,Yes,No,
```
hewanadun, line 37: Similar as above, this check should be "If MODEA2 = 1 or 3" ?

forms/uds/a2/form_a2_ivp_error_checks_mc.csv
```diff
@@ -1,26 +1,40 @@
...
+a2-ivp-c-037,037,Error,a2,I,RMMODEA2,Conformity,RMMODEA2 conforms,RMMODEA2 must be an integer between 1-2,Q0d1a. RMMODEA2 (Mode of remote A2 administration) must be an integer between 1-2,If RMMODEA2 < 1 or RMMODEA2 > 2,,,Yes,No,
+a2-ivp-m-038,038,Error,a2,I,A2NOT,Missingness,A2NOT conditionally present,A2NOT cannot be blank,"If Q0d. MODEA2 (A2 in person, remote or not completed) = 3 (not completed) then Q0d2. A2NOT (If form was not submitted, specify reason) cannot be blank",If MODEA2 = 3 and A2NOT = blank,,,Yes,No,
```
hewanadun, line 39: Don't we also need to check whether A2NOT is blank when MODEA2 != 3? (similar to checks for other variables)

hewanadun (changes requested):
@kgauthreaux 

- For optional forms such as this one, if MODE is set to "Not completed", is it OK only to validate the header variables? Or do we need to make sure rest of the form is not filled? If we need to validate that, please add another error check under MODE variable (something like _If MODExx = 3, form should not have data filled..._)

- In this file changes are not tracked correctly from the previous version, previous content is shown as deleted and everything is shown as newly added. So I cannot figure out whether there were any changes to previous checks. I assumed only thing changed was the addition of header variables, let me know if there were any other changes.

echeng06:
@kgauthreaux Also want to add Q&V csv should update `FRMDATEA2` and `LANGA2` should have their missingness column set to "Always" to reflect the rule requirements

echeng06 (changes requested):
There are two MC CSVs, I think `a2_ivp_error_checks_mc.csv` is the correct one but please rename it to `form_a2_ivp_error_checks_mc.csv` and delete the old one. Also
* `a2-ivp-m-038`: Needs to be if MODEA2 is 0, not 3
* `a2-ivp-m-040`: Needs to be if MODEA2 is 0, not 3

hewanadun (approved):

PR #190, opened 2024-08-24
Kgaut chandima file edits b1

kgauthreaux:


echeng06 (changes requested):
* `B1NOT` `short_desc` needs to be updated to `If MODEB1=3 then B1NOT cannot be blank`
* Similarly there needs to be another check for `If MODEB1 = 1 then B1NOT must be blank`
* Also see https://github.com/naccdata/uniform-data-set/issues/195

echeng06 (approved):
Updates look good!

PR #191, opened 2024-08-26
Kgaut chandima file edits b4

kgauthreaux:


hewanadun (changes requested):
@kgauthreaux 
In questions_and_vars csv, for CDRSUM, data_type is "Integer" and conformity set to "Integers 0-18, 99", but response_labels has float values such as 0.5, 1.5, etc. Please change the type to "Numeric" and fix conformity in Q&V csv. It seems the error check for CDRSUM is correct, double check on that too.

hewanadun (approved):
I renamed the Q&V file to "form_b4_ivp_questions_and_vars.csv" (last underscore was missing). Noticed this for many forms, please update your local copies of Q&V to have the correct file name.

PR #192, opened 2024-08-26
Kgaut chandima file edits b7

kgauthreaux:


echeng06 (commented):
Changes in the error checks CSV:
* `B7NOT` `short_desc` needs to be updated to match error check
* `B7NOT` needs an additional rule for if `MODE7` is 1 or 2, `B7NOT` must be blank
* `b7-ivp-m-008` should check if `MODEB7` is 1 or 3

Changes needed to the Q&V CSV based on these changes:
* `FRMDATEB7`, `LANGB7` need `missingness` columns set to `Always` as they have "cannot be blank" rules
* `FRMDATEB7` should have `conformity` account for both `mm/dd/yyyy` and `yyyy/mm/dd`

echeng06 (changes requested):
See above comment

echeng06 (changes requested):
I see you added `b7-ivp-m-036` but I think `b7-ivp-m-008` and `b7-ivp-m-011` should still be 1 or 3 since they're part of the header? Or is that a misunderstanding on my part?

kgauthreaux:
@echeng06 hm I guess I missed updating that! Uploading now with the new allowed codes 0-2 for MODEXX

echeng06 (changes requested):
* `b7-ivp-m-013` is still referencing MODEB7 = 3, needs to be = 0
* `b7-ivp-m-036` similarly needs to be if MODEB7 is 0

echeng06 (approved):

PR #193, opened 2024-08-29
Kgaut chandima file edits b5

kgauthreaux:


echeng06 (commented):
For the error checks:
* `NPIQINF` needs a missingness and conformity check ("cannot be blank" and "must be being integers 1-3)
* For error `b5-ivp-m-017`, `short_desc` is supposed to be "If `NPIQINF` is not..." instead of "If `NPIQINFX` is not..."

Changes needed in the Q&V csv based on these updates:
* `FRMDATEB5`, `LANGB5` should have `missingness` set to `Always` if they cannot be blank
* `B5NOT` `short_desc` needs to be updated to match error check

echeng06 (changes requested):
See earlier comment

kgauthreaux:
Thanks @echeng06! I uploaded the requested edits to both files!

PR #197, opened 2024-09-19
B5 plausibility checks

kgauthreaux:


echeng06 (changes requested):
* Rule `b5-ivp-p-1004` should be "and" not "or
* `b5-ivp-p-1025/26` has some typos where `ANXIETY` should be `ANXIET`
* For all the equality rules (if X is 0 then Y must be 0) we account for 0 and 1, but all these variables can also be 9 (Unknown) - do we need rules to account for this?

kgauthreaux:
@echeng06 Thanks, uploading those edits! I think we can just leave the situations where these variables can =9 on either form alone since it is conceivable that a clinician would answer differently from a participant or co-participant who might not know, or vice versa. It would result in far too many alerts that would turn out to be valid situations. So I think it's ok

echeng06 (changes requested):
`b5-ivp-p-1025` is still referencing ANXIETY instead of ANXIET in the `test_logic` column.

Also we will need to update all rules to conform to https://github.com/naccdata/uniform-data-set/issues/210

echeng06 (approved):

PR #198, opened 2024-09-20
Jculhane a4 upload

jculhane:


hewanadun (changes requested):
@jculhane 

In Q&V csv,
- In PDF MODEA4 has "Not Completed" as an option, which is not in allowed values in Q&V
- Need to add A4NOT variable
- Need to split out RXNORMID1-RXNORMID40 to separate variables since that is how those variables will be submitted
    - Update the question column to "Medication 1", "Medication 2", ...., "Medication 40" respectively

In M/C checks
- comp_vars need to be filled for checks involving multiple variables
- Update checks for MODEA4 to include "Not Completed"
- Add checks for A4NOT
- Move check a4-ivp-m-015 to ANYMEDS => If ANYMEDS = 1, then at least one of RXNORMID1 - RXNORMID40 must be present
- Add new check for "If ANYMEDS = 0, all of RXNORMID1 - RXNORMID40 must be blank"
- Check a4-ivp-c-016 needs to be split into separate variables
    - It's hard to implement this check as RXCUIs can change overtime. It seems there's a way to receive weekly updates through an [API](https://documentation.uts.nlm.nih.gov/automating-downloads.html?_gl=1*1wgn5rb*_ga*MTgxODA1ODQ1My4xNzI4NDgwNzMw*_ga_7147EPK006*MTcyODQ4NDI0MC4xLjEuMTcyODQ4NDU1Mi4wLjAuMA..*_ga_P1FPTH9PL4*MTcyODQ4NDI0MC4xLjEuMTcyODQ4NDU1Mi4wLjAuMA..#download-api), but it would take extra effort and time to implement that. 
    - So let us know if it is really necessary to check whether the entered code is a valid RXCUI or would it be sufficient to just validate whether the entered value is an integer?

jculhane:
@hewanadun 
A4 was recently changed to required at all visits. I've uploaded the revised form. 

> In M/C checks
> * Move check a4-ivp-m-015 to ANYMEDS => If ANYMEDS = 1, then at least one of RXNORMID1 - RXNORMID40 must be present
I didn't make this change, since they should be starting with RXNORMID1. Instead, I added a check for each variable to ensure it's blank if the one before is blank.

>   * So let us know if it is really necessary to check whether the entered code is a valid RXCUI or would it be sufficient to just validate whether the entered value is an integer?
Yes, this check is necessary. Checking after submission (e.g. at the freeze) will not work since at that point we'd just have to delete data.

hewanadun:
> I didn't make this change, since they should be starting with RXNORMID1. Instead, I added a check for each variable to ensure it's blank if the one before is blank.

@jculhane above doesn't work with the way the REDCap form is designed. In the REDCap form the drugs are divided into three sections and it seems users can fill any section. So RXNORMIDxx vars are not necessarily filled in sequence.

hewanadun:
Please move check a4-ivp-m-056 to ANYMEDS
In full_desc it says "Name of medication" when referring to RXNORMIDxx variables, this is kind of misleading since the variable value is expected to be the code not the name. Can you change that.

Are there any plausibility checks for A4?

jculhane:
No plausibility checks for A4

hewanadun (approved):
Looks good!

hewanadun:
@jculhane Sorry, I forgot to ask this before
In REDCap instructions it says "we recommend keeping at least a few of these fields so that you can enter "0 = No RXCUI code available" for the rare cases in which an RXCUI is not available in the RXNorm database".
So, it seems '0' is a valid response for RXNORMIDxx variables. Do we need to update PDF and error checks to reflect that?

jculhane:
Good catch. We will add this to the PDF. I've updated the var list and error checks.

PR #199, opened 2024-09-20
Upload C2 variable checklist and m/c checks

jculhane:


echeng06 (changes requested):
MC Error Checks

* Please update the filename to `form_c2_ivp_error_checks_mc.csv so we're consistent across forms
* `c2-ivp-c-002`: Date should use `/` instead of `-` to make it consistent with other forms
* `c2-ivp-c-012`: `RMMODEC2C2T` can only be 2 according to the PDF/Q&V CSV
* `c2-ivp-c-016`: This is usually just `MOCAREAS must be between 95 and 98`, since these are the ONLY values its allowed to be no matter what. `c2-ivp-c-015/17` cover the conditional cases
     * A majority of rules/fields have the same concern as above
* `c2-ivp-m-189` covers `REY1REC` being blank when `VERBALTEST` is 2, but I think we may need to further cascade this for everything dependent on `REYDREC`, since nothing constrains them when `REY1REC` is blank. Like "all of these are blank if VERBALTEST is 2/REY1REC is blank"
    * Same situation for `REYDREC`, `CERAD1REC`, `CERADJ6REC`
* `c2-ivp-m-324`: Need an inverse, e.g. if `RESPVAL is 1, then all of those fields must be blank`
* `c2-ivp-c-287/278`: `short_desc` is missing the starting "If"

echeng06 (changes requested):
On the errors MC CSV, `c2-ivp-m-189` and `c2-ivp-m-249` need `comp_vars` set, but otherwise it looks good!

echeng06:
I believe `c2-ivp-c-325` should still be under `MOCATOTS` not `MOCATRAI`

echeng06 (dismissed):

hewanadun (changes requested):
@jculhane 
Please add the check for extra C2T variables, see the comment in [C2T PR](https://github.com/naccdata/uniform-data-set/pull/204#issuecomment-2405257838)

echeng06 (changes requested):
Plausibility:
* `c2-ivp-p-1001 - 1008`: Please update `short_desc` and `full_desc` to follow the follow a "If X is ... then Y must be ..." format
* `c2-ivp-p-1010`: Can you update `short_desc` to also say that `UDSVERFC` and `UDSVERLC` must not be 95-98?

echeng06 (changes requested):
Plausibility looks good, also minor Q&V csv error:
* `MOCALANX` has an incorrect response label

echeng06 (approved):

hewanadun:
@jculhane 

- In check c2-ivp-c-022, can you change `short_desc` to 'MOCALAN must be between 1 and 3' to consistent with other checks of this type
- Does MOCALAN need another plausibility check for 1,3 (If MOCALAN is 1,3 then LANGC2C2T should not = 2)?
- Similarly, does NPSYLAN need another plausibility check for 1,3 (If NPSYLAN is 1,3 then LANGC2C2T should not = 2)?
- Please double check the allowed values, there are several places that say "must be between 0 and 77 or 95 and 98", which should be corrected to "must be between 0 and 77 or **between** 95 and 98"
- For c2-ivp-p-1011 and c2-ivp-p-1012, can you update these checks to sum to be computed only if subquestions are filled?

echeng06:
For plausibility, `c2-ivp-p-1011`: `udsverfn` is misspelled in `comp_vars` (missing s)

jculhane:
> * For c2-ivp-p-1011 and c2-ivp-p-1012, can you update these checks to sum to be computed only if subquestions are filled?

This edit led me to fix some of the m/c checks for these variables as well.

echeng06 (commented):
@jculhane Based on the update in the M/C checks, do we also need to update `c2-ivp-m-179` and `c2-ivp-m-181`? It seems a bit weird to me to restrict `c2-ivp-m-176` for `UDSVERTN` but not the other two totals. I don't know if it's better for them to be reliant on `UDSVERFC and UDSVERLC`, which direct the skipping on the PDF, or the fields they actually rely on, but I guess either works.

jculhane:
@echeng06 I think it's ok as-is. Let me know if you think something doesn't work.

hewanadun:
@jculhane there are a couple of places that still missing "**between**" in "or 95 and 98", otherwise looks good!

hewanadun (approved):

PR #200, opened 2024-09-20
Upload B3 variable checklist and m/c checks

jculhane:


echeng06 (changes requested):
Please rename the error files in lowercase for consistency. 

Errors M/C:
* `b3-ivp-m-007`: `short_desc` should be conditional; "If MODEB3 is 3 then B3NOT cannot be blank"
* `b3-ivp-c-010`: PDNORMAL should be blank or 1 (not 0 or 1). Corresponding rules dependent on PDNORMAL being 0 should be updated to PDNORMAL being blank
* `comp_vars` is missing for all rules with multiple variables

Errors P (not sure if this is complete, but if it is):
* `b3-ivp-p-1001/1002/1008` are these also comparing to B8? They only list B3 right now
* `b3-ivp-p-1007/8`: Might be my lack of context, but the and/or is not clear to me, should these just be "or"? 
* Please clean up `short_desc` for `1003-1006` for consistency/succinctness (e.g. don't need the extra stuff in parentheses) 
    * In general can all be reworded to "If X, Y, and Z are 0, then A cannot be 1-3"
* Neds to be updated to follow https://github.com/naccdata/uniform-data-set/issues/210

jculhane:
@echeng06 I just noticed there are two "013" m/c checks. Should I make one #150 instead?

echeng06:
@jculhane Yes that's fine, thanks!

echeng06 (changes requested):
* On M/C errors, everywhere it says "If PDNORMAL is 0" should be changed to "If PDNORMAL is blank", since it can only be blank or 1
* Some of the old files got leftover, please clean up on the next upload
* For plausibility errors, some of the changes results in variables from B8 being the `var_name`, either we need to move these rules to B8 or switch the order of the rules themselves so they're only under B3 variables
    * e.g. `b3-ivp-p-1003-1007`

echeng06 (changes requested):
Looks like the question and vars CSV got deleted, otherwise the changes look good!

echeng06 (dismissed):

hewanadun (changes requested):
@jculhane In the PDF it says if PDNORMAL is checked (i.e. =1), all items will be default to 0 in the database. So I think it is okay if they check PDNORMAL and still select the 0 option for the questions.
Current checks says **"If PDNORMAL is 1 then [var name] must be blank"**, I think these should be changed to **"If PDNORMAL is 1 then [var name] must be 0 or blank"?** WDYT?

echeng06 (changes requested):
* `b3-ivp-p-1012` This is okay but I think it'd make more sense here to say we should calculate if they're not 8, since their actual maximum is 4. Otherwise changes look good!

hewanadun:
@jculhane can you please delete the I4 plausibility file from this branch and submit it as a separate PR in a new branch. Thanks!

hewanadun (approved):

echeng06 (approved):

PR #201, opened 2024-09-21
Upload A4a variables, m/c, and plausibility

jculhane:
No plausibility checks for this form


echeng06 (changes requested):
Please update filenames to be all in lowercase just for consistency, thanks!

* Please update `comp_vars` for all rules missing it, re: https://github.com/naccdata/uniform-data-set/issues/210
* `a4a-ivp-c-002`: Some of the entries are using dashes `-` instead of slashes `/` for the delimiter, please update to `/` for consistency
* Everything that relies on TRTBIOMARK being 0 should be updated to TRTBIOMARK being 0 or 9, since both options end the form (and force all other variables to be blank)
* Same thing as above for ADVEVENT
* `a4a-ivp-m-058` - `a4a-ivp-m-060`: I don't think it's right for these to be dependent on `TRTBIOMARK` since its the second one, it's more correct to say these should only be filled out if one of the primary drug target (2) is filled out. 
    * At the very least the ones that are reliant `TRTBIOMARK` being 1 should be changed to be reliant on the primary drugs instead. The ones that are reliant on it being 0/9 for blankness are okay to leave as is
    * This goes for all other trials on page 2 (targets 3-8) 
    * On that note, for `a4a-ivp-m-024` - `a4a-ivp-m-044` it's fine that they are reliant on `TRTBIOMARK` since its the first one, but if we wanted it to make it consistent with the others I think they should also be reliant on the primary drug targets (1)

jculhane:
I _think_ everything above is taken care of - please let me know if I've missed something.

echeng06 (changes requested):
Minor thing but a lot of typos:
* In MC errors, "intgeger" instead of "integer"
* Q&V, This is actually in a lot of forms, not just this one, sorry I'm only noting now, but
    * "Intgers" instead of "integers" 
    * "Condtional" instead of "Conditional"
    * For these I don't think we really need a PR on every form, we can fix these typos at the end 

Errors:
* Can we expand the "if any drug targets are 1 then TRTTRIALX cannot be blank" (e.g. `a4a-ivp-m-058, 082, etc.`) to include the rest of the (required) values as well? Otherwise none of them have rules for needing to be present if the drugs were filled out
    * i.e. "If any of (TARGETAB2, TARGETTAU2, TARGETINF2, TARGETSYN2, TARGETOTH2) is 1 then all of (TRTTRIAL2, STARTMO2, STARTYEAR2, ENDMO2, ENDYEAR2, CARETRIAL2) cannot be blank"
        * I don't think we've defined a standard in #210 for this, but after talking with Chandima, we were thinking **"If any of A,B,C is 1 then X, Y, Z cannot be blank" should be written under A**, so in this case under `TARGETABX` instead of `TRTTRIALX`
    * OR each of the variables will need to duplicate this rule, which will just result in more error checks but a "simpler" rule - I'm fine either way
    * Chandima also pointed out it's not clear from the form alone that all those columns/variables are required to begin with if the drug targets are filled out - I was basing it off the fact that the first row does require all of them if Q1 was 1 (yes). If they are not required then we don't actually need this change
* The above set of rules also have some typos/inconsistencies:
    * In `short_desc` all are still referencing `TRTBIOMARK`
    * Many columns are saying = X instead of = 1

echeng06:
Apologies a few more things on the errors:

* First error (a4a-ivp-m-001) is referencing `FRMDATEA1` instead of `FRMDATEA4A`
* In general any variable with "A4a" should have everything capitalized to "A4A"
* `a4a-ivp-m-027` should be under `TRTBIOMARK` instead per #210

echeng06 (changes requested):
Last nitpick is to make the filenames all lowercase (e.g. a4a instead of A4a) but otherwise the changes look good!

echeng06 (approved):

hewanadun:
@jculhane sorry for requesting a change after this PR is closed.

As Emily mentioned above
> * we were thinking **"If any of A,B,C is .... then X, Y, Z ..... " should be written under A**

So could you please move the checks in the format
If TARGETABx, TARGETTAUx, TARGETINFx, ...... are blank then TARGETABx+1, TARGETTAUx+1, TARGETINFx+1, ...... must be blank to variable **TARGETABx**. (Currently they are written under **TARGETABx+1**)
For example, check a4a-ivp-m-057 should be written under **TARGETAB1** instead of **TARGETAB2**

(Just want to note that the checks in the format of a4a-ivp-m-058 are written under the correct variable, no need to change those)

echeng06:
@jculhane Apologies I am also seeing another thing, `a4a-ivp-m-228` should be under `ADVEVENT` instead of `ARIAE`

PR #202, opened 2024-09-22
Upload B9 form, m/c checks, and plausibility checks

jculhane:


echeng06 (changes requested):
Q&V CSV:

* `FRMDATEB9` conformity should be `mm/dd/yyyy` or `yyyy/mm/dd`

Errors MC

* `b9-ivp-c-035`: `COGORI` is misspelled (swapped G and R)
* `b9-ivp-m-039`: `short_desc` should be dependent on `DECCLCOG`, not `DECCLIN`
* `b9-ivp-m-085`: should list out the actual variable names (and be 12a-12f not 14) and put under `comp_vars`
* There should also be an inverse of the above rule (if all are 0 (No) or 9 (Unknown) this should be blank)
* `b9-ivp-m-091` and `b9-ivp-m-094`: There should be inverses of these rules (if `BEVHALL` is 0 or 9 then these should be blank)
* `b9-ivp-m-100` and `b9-ivp-m-103`: Same as above
* `b9-ivp-m-112`: should list out actual variable names (and be 12h-12k not 14) and put under `comp_vars`, and also reference `PSYCHAGE` not `AGITAGE`
* Also should be an inverse of the above (if all are 0 or 9 should be blank)
* `PERCHAGE` needs 2 rules to check against 12m-12r, list out all variables and put under `comp_vars`, similar to the above 2 points
* `b9-ivp-m-150` and `b9-ivp-m-153`: should be 0 or 9, not just 0
* `BEREMAGO` and `BEREMCONF` need rules to be blank if `DECCLBE` (or `BEREM`, doesn't matter) is blank
* `BEOTHRX` needs a rule to be blank if `DECCLBE` (or `BEOTHR`, doesn't matter) is blank. Could also change `b9-ivp-m-159` to be "If BEOTHR is NOT 1 then BEOTHRX must be blank", since that implicitly works, but think there should just be a separate rule so its clear since its really a dependency on DECCLBE
    * I say the above because rules like `b9-ivp-m-063`, `b9-ivp-m-164`, and `b9-ivp-m-196` handle that case implicitly (by saying "not 4"); whichever way we go we should be consistent
* `b9-ivp-m-189`: Should list out actual variable names (and be 15a-15g, not 17) and put under `comp_vars`
* Above also needs an inverse (if all or 0 or 9 should be blank)

Errors Plausibilty:

* `b9-ivp-p-1008`: Should this be "and" not "or"? 
    * I also really don't know how to write this rule on our end, particularly because `FRMDATEB9` will be in a string-date form so isn't readily "comparable" to the other values. Maybe @hewanadun has some ideas?

hewanadun:
@jculhane 
Please move check b9-ivp-p-1008 to Form A1 variable BIRTHYR (error code also needs to be changed)

@echeng06 we'll need a custom rule to handle this

jculhane:
@hewanadun moved

echeng06 (changes requested):
Errors MC:
* `b9-ivp-m-085-86` should be under `BEHAGE`, not `BEAPATHY`
* `b9-ivp-m-112-113` should be under `PSYCHAGE` not `BEVHALL`
* `b9-ivp-m-143-144` should be under `PERCHAGE`, not `BEDISIN`
* `b9-ivp-m-092, 95, 101, 104`: Basically anywhere that is "0 or 9" I think we should change to "not 1" since this will cover the other variable being blank case (which is missing otherwise) and be consistent with all the other rules
* `b9-ivp-c-191` has `comp_vars` filled out, please remove

Plausibility checks look good to me!

echeng06 (approved):

hewanadun:
@jculhane another small typo

In b9-ivp-p-1015 **test_name** uses the wrong variable DECCLCOG instead of DECCLIN

echeng06 (approved):

PR #205, opened 2024-09-23
Upload B6 form and error checks

jculhane:


echeng06 (changes requested):
There are some extra files in the folder, please clean up on the next upload!

Errors MC:
* Please update `comp_vars` for all rules that compare multiple variables (pretty much just the header stuff at the bottom)
* `b6-ivp-m-044`: Should check if MODEB6 "is 1 or 3" not just 1
* `b6-ivp-m-046`: B6NOT is conditional, rules need to be updated to reflect this (change to "If MODE6 is 3 then B6NOT cannot be blank" and then need another rule for the inverse)
* `b6-ivp-c-047`: B6NOT must be 95-98, not 94-98

Still looking at the plausibility checks, they look fine aside from `b6-ivp-p-1004` looking difficult to implement, but I'll get back to you on that.

echeng06 (approved):
Looks good, thanks! The plausibility (GDS) checks also look good, we'll just need to update some code on our end to support it.

hewanadun:
@jculhane it seems we are missing a plausibility check for GDS
**IF NOGDS=0 then there must be at least 12 questions with valid scores (0 or 1)**
Or is it okay to compute partial GDS even if the number of questions with valid scores < 12? (If so b6-ivp-p-1004 needs to be updated to clarify that)

jculhane:
Added that plausibility check. I also edited the "test_logic" for 1002 as I don't think it was correct before.

PR #207, opened 2024-09-27
B8 error checks m/c and p

kgauthreaux:


echeng06 (changes requested):
There seem to be some extra files in here, please clean up and make sure the filenames are consistent (e.g. question and vars needs all underscores)

Errors P:

* `b8-ivp-p-1001/1009` correspond to M/C errors as `b8-ivp-m-125/126` - is there a reason 2 are plausibility and the other 2 aren't? Also might be worth combining them to instead say at least one of them must be 1, 2, or 3
* `b8-ivp-p-1001-1009` needs to be updated to follow https://github.com/naccdata/uniform-data-set/issues/210 convention (swapping the order of the condition fine)

Errors MC:

* The header in the PDF is not like the other forms I've seen and as such doesn't have stuff like `RMREASB8` and `RMMODEB8` - does the PDF or rules need to be updated? 
* None of the rules except `b8-ivp-m-003` take into account `NEUREXAM` being 2 or 3 - just want to confirm the correct behavior here is to allow fields to be blank or filled in without being constrained to it, since that is what's happening here as a result (I believe it is just want to double check to make sure we're not missing anything!)

echeng06:
@kgauthreaux Just bumping this one for visibility since it's been sitting idle for a while.

kgauthreaux:
@echeng06 thanks! I was waiting until I start my review of the iv4 plausibility to get to B8! and okay I combined those issues into b8-ivp-p-1001/1009 and deleted b8-ivp-m-125/126. The updated PDF has not been uploaded but it does have the headers - although note the change to the questions and vars list for RMMODEB8 - the only acceptable remote mode for this exam is 2 (Video). 

And we don't have to take NEUREXAM into account for the conditionally present/blank since there is not a set list of Qs that count as a "focused or partial exam" - to my knowledge it varies by the examiner and participant on the day. So we can ignore it, and expect a lot more "not assessed" 8s throughout the form for those exams than usual.

echeng06 (changes requested):
There are 2 Q&V CSV files here, please remove the outdated one

M/C:
* `b8-ivp-m-112`: Can simplify to just say "If GAITFIND is not 7 then"
* In general I wonder if we can simplify all the
    * "If NEUREXAM is 0 or NORMNREXAM is 0 or SOMEVAR is 0 or 8 then VAR must be blank"
    * to "If SOMEVAR is not 1 then VAR must be blank"
    * Since SOMEVAR being blank implies "NEUREXAM is 0 or NORMNREXAM is 0"? 
    * But also fine to leave as is if we don't want to bother changing it, both work. I've only recently become super conscious of the above situation, so we might not be doing this in all forms atm. 
 
Plausibility:
* `b8-ivp-p-1015`: `short_desc` should be "or" not "and" on the 1-4 variables

kgauthreaux:
@echeng06 okay I tried to simplify all of those M&C checks but may have missed some so just let me know!

echeng06 (dismissed):

hewanadun (changes requested):
Some of the checks in b8-ivp-p-1016 - b8-ivp-p-1021 are duplicated in B3 b3-ivp-p-1009 - b3-ivp-p-1011, there are two options
    - Remove b3-ivp-p-1009 - b3-ivp-p-1011 from B3
    - Move the checks that are only in B8 also to B3 (logic is simpler if we implement them in B3) and remove b8-ivp-p-1016 - b8-ivp-p-1021 from B8

kgauthreaux:
@hewanadun sorry I am a bit confused by your comment. Is a third option to just remove the duplicate checks from B8? I am seeing duplicates for b8-ivp-p-1016, b8-ivp-p-1018, and b8-ivp-p-1020 in B3

hewanadun:
> @hewanadun sorry I am a bit confused by your comment. Is a third option to just remove the duplicate checks from B8? I am seeing duplicates for b8-ivp-p-1016, b8-ivp-p-1018, and b8-ivp-p-1020 in B3

Yes, that is fine too if they don't need to be in one place.

PR #174, opened 2024-05-21
Kgaut chandima file edits a1

kgauthreaux:

PR #208, opened 2024-09-27
kgaut-upload-a2-plausibility

kgauthreaux:


echeng06 (changes requested):
* `a2-ivp-p-1001`: Seems like an incomplete rule? Is this supposed to be "If INRELTO is 1 then INLIVWITH cannot be 0", which also seems kind of invalid (they don't necessarily have to live together)? It would also contradict `a2-ivp-p-1004`
* `a2-ivp-p-1002`: `LIVSITUA` needs to be explicitly mentioned in `short_desc`
* `a2-ivp-p-1009`: Typo in `comp_vars`, should be DECCOGIN
* `a2-ivp-p-1010`: Should this be "If INMEMTROUB is 2,3,4,5" instead of "1,2"? 
* `a2-ivp-p-1011`: Similarly I think this should be 4,5, not 1,2
* For all, please update the rules to follow the convention mentioned here, e.g. the "then B" should be the one listed under `var_name` so need to switch the order: https://github.com/naccdata/uniform-data-set/issues/210

hewanadun:
> For all, please update the rules to follow the convention mentioned here, e.g. the "then B" should be the one listed under `var_name` so need to switch the order: [Error checks involving multiple variables #210](https://github.com/naccdata/uniform-data-set/issues/210)
> * That or rules might need to be moved around to different forms

@echeng06 @kgauthreaux If there are only two variables involved in the check, it can be written under either variable, please see https://github.com/naccdata/uniform-data-set/issues/210#issuecomment-2397327749

kgauthreaux:
@echeng06 a2-ivp-p-1001 was an alert in UDSv3 so we copied it over to UDSv4. it's just an alert so they can confirm it if it's the case. Then the rule in a2-ivp-p-1004 can apply

hewanadun:
@kgauthreaux can you please move a2-ivp-p-1003 and a2-ivp-p-1004 to Form A1 - LIVSITUA. 
We requested some [other changes ](https://github.com/naccdata/uniform-data-set/pull/174#issuecomment-2433494269)in A1 too, above can be done together with those.

hewanadun (approved):

hewanadun:
@echeng06 I think this one is ready to merge. Please double check the A2 rule definitions and approve this PR.

echeng06 (approved):

PR #211, opened 2024-10-21
Upload E3F var list and m/c

jculhane:


echeng06 (changes requested):
* `e3f-ftldivp-m-170` is missing `FTDCBFOA` in `comp_vars`

Looks good otherwise!

echeng06 (approved):

hewanadun:
@jculhane 
For the string variables that say "Any text or numbers, but cannot include single quotes (’), double quotes (“), ampersands (&) or percentage signs (%)" in Q&V csv (e.g. FTDMRIOS, FTDFDGOS, etc.), do we need to specifically check if this condition satisfied? If so, please add a conformity check for each for those variables.

jculhane:
No, we don't - the module Q&Vs are pulled directly from the current (v3) DEDs, and these were requirements in the current SAS system.

PR #212, opened 2024-10-21
Upload C3F var list, m/c, and plausibility

jculhane:


echeng06 (changes requested):
Looks like E3F was uploaded instead of C3F

echeng06 (approved):

PR #213, opened 2024-10-21
Upload C1F var list and m/c

jculhane:


echeng06 (changes requested):
M/C errors:
* Lots of typos of repeating "a an" or "a a" in `short_desc`
* `c1f-ftldivp-c-026`: `test_logic` should be comparing to 20, not 15
* `c1f-ftldivp-c-028`: Should be 0-8 not 0-1/15 for several columns
* `c1f-ftldivp-c-053`: `full_desc` and `test_logic` should be comparing to 20, not 10
* `c1f-ftldivp-m-057`: `short_desc` should be "should be blank" instead of "should blank"
* `c1f-ftldivp-c-061`: Should say "number" instead of "integer" 
     * `c1f-ftldivp-c-063`: Similarly can change this to say "integer" instead for consistency 
* `c1f-ftldivp-c-066`: `full_desc` and `test_logic` should be comparing to 37, not 10

echeng06 (approved):

PR #214, opened 2024-10-23
Update B5 MODEB5 variable conformity

kgauthreaux:
Updating MODEB5 variable conformity based on discussion for FTLD/LBD module vars


echeng06 (changes requested):
* `b5-ivp-m-013`: Should be if MODEB5 is 0, not 3
* `b5-ivp-m-015`: Should be if MODEB5 is 0, not 3

echeng06 (approved):

Issue #210, opened 2024-09-28
Error checks involving multiple variables


hewanadun:
@jculhane @kgauthreaux 

Noticed this in A5D2 error checks but reporting as a general issue since this is relevant to all forms.

1. For error checks involving multiple variables or forms, "comp_vars" column should include **all** the variables involved in the check and "comp_forms" column should include **all** the forms involved in the check. For example,
      - In check a5d2-ivp-m-007, comp_vars column should have "MODEA5D2, RMREASA5D2", it's currently blank
      - In a5d2-ivp-m-137 and a5d2-ivp-m-352, HEADIMP is missing in the comp_vars list
      - In a5d2-ivp-m-169 and a5d2-ivp-m-354, DIABETES is missing in the comp_vars list
These are only couple of examples, there are many checks where comp_vars are missing or incomplete.

2. When choosing the variable to report the error, please follow the following convention.
    - If the check is in format, **"If A is .... then B must be ...."** => write the error under **variable B**.
        - Most of the error checks belong to this format and the variables RT using to report the error is correct for this type of checks (e.g. a5d2-ivp-m-007, a5d2-ivp-m-029, etc.)
    - If the check is in format, **"If at least one of P, Q, R, S, T  must be ...."** => write the error under **variable P**.
        - Some examples are a1-ivp-m-063, a1-ivp-m-076, these checks are also reported under the correct variables
    - If the check is in format, **"If X is .... then (at least one of |  none of | all of | some of | etc.) P, Q, R, S, T  must be ....."** => write the error under **variable X**.
        - _**This is the case where RT error checks are reported under the wrong variables**_. Some examples
            - a5d2-ivp-m-137 and a5d2-ivp-m-352 - these checks should be moved to variable HEADIMP (currently reported under IMPAMFOOT)
            - a5d2-ivp-m-169 and a5d2-ivp-m-354 - these checks should be moved to variable DIABETES (currently reported under DIABINS)
        - Only the var_name column needs to be updated, no changes required in rest of the columns including the error_code. Above are only couple of examples, there are more similar cases in A5D2 and maybe in other forms. Please check and fix.
        - Just want to mention that checks in B5 plausibility (b5-ivp-p-1003, b5-ivp-p-1004, b5-ivp-p-1023, b5-ivp-p-1024, etc) belongs to this format and reported under the correct variable. So, please follow the same convention for any other checks in the above mentioned format.

Requesting these changes as it helps us to generate more meaningful error report.
[...]

PR #215, opened 2024-10-23
Upload A5D2 I4 plausibility

jculhane:


echeng06 (changes requested):
* `a5d2-ivp-p-1001`: Should be BIRTHSEX=2 not 1 for several columns
* `a5d2-ivp-p-1003`: Typos, should be ALCDRINKS not ALDRINKS
* `a5d2-ivp-p-1007`: Should be under RBD, and on the IVP one we also swapped the order, so we should probably do the same here for consistency (and will be consistent with the other rules as well)
* `a5d2-ivp-p-1020`: Typos, should be OTHANXD not OTHAND
* Error codes need to say `i4vp` not `ivp` 
* Do PDOTHR and CBTIA also need a comparison between UDSv3 and v4?
* `a5d2-i4vp-p-1036` and `a5d2-i4vp-p-1038` are the same, was the latter meant to be CBTIA?

jculhane:
Apologies for somehow starting with an old version of IVP for these. Fixed!

echeng06 (approved):

hewanadun:
@jculhane  As per Teams discussion, please update any equality checks against UDSv3. See section "Exclude Unknown or N/A codes when comparing with a previous visit" in https://github.com/naccdata/uniform-data-set/issues/210#issue-2554408480

PR #216, opened 2024-10-23
Upload B4 I4 plausibility

jculhane:


echeng06 (changes requested):
* `b4-i4vp-p-1002` and `b4-i4vp-p-1003`: Can you update these to follow "If ... then ..." format just for consistency?
* `b4-i4vp-p-1004` and `b4-i4vp-p-1005`: For `short_desc` please remove the stuff in parentheses 
    * `b4-i4vp-p-1005` should also say "is not <4" instead of "should not <4"
* `b4-i4vp-p-1003`: Under `test_logic` CDRGLOB should be 99, not 9
* `b4-i4vp-p-1019`: Reword `short_desc` to be something more like the difference between the two visits should not be more than 1, instead of an if statement

hewanadun:
@jculhane We need some changes in both I4 and IVP plausibility checks
 - In b4-i4vp-p-1001, update the check to compute sum only if all subquestions have a valid response. 
     - Similarly, IVP check b4-ivp-p-1001 needs to be updated too.
- I added a new section to https://github.com/naccdata/uniform-data-set/issues/210#issue-2554408480. Please see **"Cross form error checks involving C2 or C2T forms"**
    - Update b4-i4vp-p-1004 and b4-i4vp-p-1005
    - Similarly, in IVP checks b4-ivp-p-1004 and b4-ivp-p-1005 need to be updated too.
 - In b4-i4vp-p-1019, update `sort_desc` to `CDRGLOB should not decrease by 1 or more points from CDRGLOB[UDSv3]`

PR #220, opened 2024-10-26
FTLD - C1F plausibility checks

kgauthreaux:


echeng06 (changes requested):
* `c1f-ftldivp-p-1001` is listed twice, can you update `error_code` to match `error_no`?

hewanadun (changes requested):
- In 1002 - 1006: Sum or ratio should be computed only if variables involved have valid values (i.e. should exclude 95-98 or blank). Please update the checks accordingly

echeng06:
@kgauthreaux to add onto Chandima's comment, 1006/FTDRATIO will need to account for the divide by 0 case - 1006 could say "if they're 1-16" instead of the usual "0-16" but we also need to decide what to do if one or both of the variables are 0.

hewanadun:
@kgauthreaux Checks 1002 to 1005 needs to be updated too (only 1006 was updated)

> * In 1002 - 1006: Sum or ratio should be computed only if variables involved have valid values (i.e. should exclude 95-98 or blank). Please update the checks accordingly

kgauthreaux:
@hewanadun okay sorry I thought those sums were taken care of by the missingness rules. I will re-upload.

echeng06 (changes requested):
* `c1f-ftldivp-p-1007` has a "non-sero" typo
* `c1f-ftldivp-p-1002 - 1006`: Sorry to be nitpicky but can we change short_desc to just say "are not equal to 95-98" instead of "non-missing (not equal to 95-98)"
        * `1006`: For this one we can just say if the two are 1-16
        * "If FTDNOUNC and FTDVERBC are 1-16 then FTDRATIO must equal FTDNOUNC / FTDVERBC"

kgauthreaux:
@echeng06 okay I changed c1f-ftldivp-p-1006 but I will keep the rest as is because they still are easy enough for me to understand

echeng06 (approved):

hewanadun (approved):

PR #223, opened 2024-10-27
Upload revised B6 IVP plausiblity

jculhane:


hewanadun (changes requested):
@jculhane two minor changes,
In check b6-ivp-p-1005, add GDS to `comp_vars`
In check  b6-ivp-p-1004, can you please move the details of the algorithm to `test_logic` and make `short_desc` and `full_desc` simpler (I don't think we would be able to show that long paragraph in error report UI)

jculhane:
For 1005, test_name and var_name should actually have referred to NOGDS.

hewanadun:
@jculhane 

- I agree that check 1005 better reported under `NOGDS`, same is for check 1002. However, since this is a special computation, we had to group the logic together in one function (we cannot validate 1003 and 1004 without validating 1001, 1002 and 1005 first). So we need to bend the rules and have these errors written under one variable. So could you please move 1005 back to `GDS`.
- Just noticed that description for partial GDS calculation (1004) is not complete. It ends at "Since the decimal portion of this value is", could you specify what is supposed to happen then? Since GDS is defined as Integer, do we need to round off?
- In m/c checks noticed that `test_logic` uses `NACCGDS` (should be changed to `GDS`)

jculhane:
I reviewed the form, v3 DED and v4 REDCap and realized that the partial GDS computation is done by NACC on the back end (i.e., in a NACC derived variable called NACCGDS). Therefore, I removed check 1004, and revised 1003 to reflect that GDS should be the simple sum of all valid sub-questions. I wasn't sure about the wording on this check (it should only add 0s and 1s from sub-questions, not 9s).

hewanadun:
@jculhane Please add `GDS` to 1005 `comp_vars` list (it should be similar to 1002)

@echeng06 I updated the validator code and definitions to reflect above changes.

hewanadun:
@jculhane sorry, I missed this earlier, since B6 is an optional form we need to add the check "If MODEB6 = 0, form should not have data filled"

echeng06 (changes requested):
* `b6-ivp-m-049`: Should be under MODEB6 not MODEXX

echeng06 (approved):

hewanadun (approved):

PR #224, opened 2024-10-27
Upload FTLD A3a var list and m/c

jculhane:


echeng06:
@jculhane Sorry I initially missed this, for `3a-ftldivp-m-001` `FRMDATEA3A` is misspelled; but it can be fixed in another PR, particularly if there is a plausibility one coming up.

PR #228, opened 2024-10-27
LBD - Upload v3 B1L missingness & conformity

kgauthreaux:


echeng06 (changes requested):
Q&V CSV:
* Please change all instances of data type of "Num" to "Integer"
* `LBSSCLOT` type should be a string, unless it really can only be a character (if so, we also need to update some code to support char validation)

Errors CSV:
* `short_desc` has a typo of "a an integer..." in a bunch of rules, please clean up

echeng06 (approved):

hewanadun (dismissed):

echeng06 (changes requested):
* `b1l-lbdivp-c-002` should be under FRMDATEB1L not INITIALSB1L

hewanadun (approved):

echeng06 (approved):

PR #232, opened 2024-10-28
Upload FTLD E3F plausibility

jculhane:


echeng06 (changes requested):
**EDIT: I guess I might have been looking at an old version, similar to D1a - if the new one is correct then disregard**

* `e3f-ftldivp-p-1002`: I think this might need to be changed to if IMAGINGDX must be 2,3
* `e3f-ftldivp-p-1003`: Similarly should be if IMAGINGDX must be 0,1

Since this is the definition for IMAGINGDX:
```
0, No|1, Yes, only PET/SPECT imaging was used|2, Yes, only MR imaging was used|3, Yes, both PET/SPECT and MR imaging were used
```

jculhane:
I waited to upload edits b/c we were waiting on updates to D1b from the CTF. Those were finished today, and there was no change to IMAGINGDX, so I made the two edits above.

echeng06 (approved):

PR #233, opened 2024-10-29
Upload FTLD C2F

jculhane:


echeng06 (changes requested):
Errors MC:
* The error codes start at 001 for the headers and then restart again at FTDCPC2F, can you move the headers to the end with new error codes
* `c2f-ftldivp-m-001 FTDCPC2F` makes all of the rules checking "If FTDCPC2F is between 95 and 98, then X must be blank" redundant, do we need all of them? 
* The last 4 totals variables (`FTDSNTOT`, `FTDSNTBS`, `FTDSNTOS`, and `FTDSNRAT`) need to be blank if `FTDCPC2F` is 95-98, so either need their own rules or get appended to `c2f-ftldivp-m-001 FTDCPC2F`

echeng06:
M/C looks good! For plausibility

* c2f-ftldivp-p-1012 and 1013 contradict each other in the case that all values are 0 or all values are 1, so I think 1012 basically needs to be the opposite of 1013 and say "cannot be 88.88 if they're all valid AND they're not all 0 AND they're not all 1"

jculhane:
The guidance specifies "If the subject answered the C2F questions all yes or all no, then these answers are considered invalid. Please select the most appropriate reason code in the “For clinic use only” section and leave the rest of the form blank." So they are not allowed to record all 1s or all 0s. I added a check for this. Can we leave 1012 as-is since the contradicting case is taken care of in a different error check, or not?

echeng06:
@jculhane Sounds good! If anything the new rule means we should simplify 1013, but otherwise that works for me.

jculhane:
You're right - 1013 is fixed.

echeng06 (changes requested):
1013's `full_desc` should say "must be 88.88", otherwise change looks good!

echeng06 (dismissed):

hewanadun:
- Typo in Q&V FTDSNRAT `response_labels`, should be 88.88
- Plausibility checks 1002 - 1013 should be evaluated only if FTDCPC2F is blank, waiting for the decision on whether we can add MODEC2F to make this logic simpler. Otherwise, these checks needs to be updated.

hewanadun (changes requested):
@jculhane @echeng06 If we add MODEC2F, DED also needs to be updated (which seems to be out of the question at this point). So, we have to go with the complex rules.

@jculhane Please update the plausibility checks as follows,
- 1002: If FTDCPC2F is blank and all subquestions have valid scores, then FTDSNTOT must equal 22 minus the sum of ......
- 1003: This check can be removed (with the above update 1003 is covered by 1002, I don't think we need to specifically check for 88)
- 1004: If FTDCPC2F is blank, then FTDSNTOT should equal 88 if any of .....
- Similar changes required in 
    - FTDSNTBS checks 1005 - 1007 
    - FTDSNTOS checks 1008 - 1010
    - FTDSNRAT checks 1011 - 1012
    
**Note:** As this is a major change, we'll have to regenerate rule definitions, so if you prefer, it is OK to re-number the error codes after removing the redundant checks.

jculhane:
@hewanadun "In 1004, can FTDSNTOT left blank too or only 88 is accepted? (similar questions for 1010 and 1012)"

These can only be blank if FTDCPC2F is not blank. I fixed the var list to clarify this.

hewanadun (approved):

jculhane:
@hewanadun I just opened a new PR for this [#311](https://github.com/naccdata/uniform-data-set/pull/311)

PR #203, opened 2024-09-22
Upload B1 form and plausibility checks

jculhane:

PR #234, opened 2024-10-30
Upload A3 I4 plausibility

kgauthreaux:


echeng06 (changes requested):
* `do_in_redcap` and `in_prev_versions` not filled out for bottom rules
* Please follow the [convention discussed here](https://github.com/naccdata/uniform-data-set/pull/203#pullrequestreview-2343909839) for writing UDSv3 vs UDSv4 variables - I've also updated #210 to document this

hewanadun:
@kgauthreaux 
In a3-i4vp-p-1044 and a3-i4vp-p-1045, remove V1/V2 from `test_logic` and `comp_forms`

echeng06 (changes requested):
* `a3-i4vp-p-1038 - 1045` Please update to follow the VARIABLE[UDSv3] notation for all columns, and remove mentions of UDSv4
    * e.g. for 1038 `short_desc`:  MOMYOB should be equal to MOMYOB[UDSv3] at the previous visit where MOMYOB[UDSv3] is not blank
* `a3-i4vp-p-1040`: The "UDSv**3**" is missing in `test_name`
* `a3-i4vp-p-1041`: Still mentioning UDSv2 in `test_logic`

kgauthreaux:
@echeng06 I thought the [UDSv3] thing was just for the test_logic column. The short_desc and full_desc are for the ADRCs to understand the error more and I think the way it's written is more informative than the way it would be written for the test logic.

echeng06:
@kgauthreaux I think we've also been updating the descriptions in other forms but I'll defer to @hewanadun if it's strictly necessary.

hewanadun:
@kgauthreaux We have been using `VARIABLE[UDSv3]` format in `short_desc` and `test_logic`. Those columns are used by rule definition generation scripts Emily wrote. We are only displaying `full_desc` in error report, that column needs to be more descriptive/informative.

kgauthreaux:
@hewanadun @echeng06 okay, I uploaded an updated version with the short_desc column in that format

echeng06 (approved):

hewanadun:
@kgauthreaux As per Teams discussion, please update any equality checks against UDSv3. See section "Exclude Unknown or N/A codes when comparing with a previous visit" in https://github.com/naccdata/uniform-data-set/issues/210#issue-2554408480

kgauthreaux:
@echeng06 @hewanadun I wanted to check on this before I open a new branch and upload: 

The SIBS "missing" code is 77 which means "adopted, unknown." I feel like if they answered 77 in UDSv3 it would be okay to keep the rule as it is because it should also =77 in UDSv4? At least logically - though idk if they would ever put 77 and then learn about siblings through their adopted parents and then try to answer the question with a code between 1-20? Seems like a rare enough situation that in that case they would just verify the alert. I'd like to hear your thoughts on it

PR #236, opened 2024-10-30
FTLD - Upload E2F missingness & conformity

kgauthreaux:


echeng06 (changes requested):
* Error codes should be `ftldivp` not `ivp`
* `e2f-ivp-m-007` and `e2f-ivp-m-010`: `short_desc` typos, should be "If FTDSMRI is 0" not "If FTDSMMO is 0"
* `e2f-ivp-m-017`: `short_desc` uses a period instead of comma
* The following all need a corresponding rule to say that that field should be blank if the field it depends on is blank (these are basically all the text fields)
    * `e2f-ivp-m-015` for FTDSMDIS
    * `e2f-ivp-m-019` for FTDSMADV
    * `e2f-ivp-m-023` for FTDSMMAO
    * `e2f-ivp-m-028` for FTDSMFSO
    * `e2f-ivp-m-046` for FTDFDDID
    * `e2f-ivp-m-050` for FTDFDADV
    * `e2f-ivp-m-054` for FTDFDMAO
    * `e2f-ivp-m-073` for FTDAMDID
    * `e2f-ivp-m-077` for FTDAMLIO, also relies on not FTDAMLIG being 3, not FTDFDADN being 1
    * `e2f-ivp-m-081` for FTDAMADV, also relies on FTDAMADN not FTDFDMAN 
    * `e2f-ivp-m-085` for FTDAMMAO
    * `e2f-ivp-m-104` for FTDOTANS
* Similarly needs a rule to be blank if FTDSMRI is 0
    * `e2f-ivp-m-024` for FTDSMMAM
    * `e2f-ivp-m-055` for FTDFDMAM
    * `e2f-ivp-m-086` for FTDAMMAM

I _think_ that's all of them but in general please ensure all the freeform text questions have both conditions accounted for!

kgauthreaux:
@echeng06 Okay I added them as checks 105-119 since it looks like the RT who wrote these forgot them. I made it so that if the main gate question =0 or if the 'other' option is not checked then it must be blank

echeng06 (changes requested):
* `error_no` needs the leading 0s
* `e2f-ftldivp-m-010`: Still has the "FTDSMMO is 0" typo in `short_desc`
* `e2f-ftldivp-m-105 - 119`: I think its sufficient to just check if the "other" option is not checked, since there's no situation where it's checked and the main gate question is 0 (that situation is disallowed by other rules), but we can also leave as is if we don't want to change it. 
    * I've started generally requesting this kind of simplification in other places, I think it makes things more readable and is simpler to evaluate but let me know if we need it to mention both otherwise
    * `e2f-ftldivp-m-110 - 112` should also say "ne" instead of "="

echeng06 (changes requested):
* `e2f-ftldivp-m-010`: `var_name` should be FTDSMYR
* `e2f-ftldivp-m-119`: Can still be simplified
* `e2f-ftldivp-m-104 - e2f-ftldivp-m-118`: Not sure what happened but it looks like half the columns shifted up one so they no longer match `var_name/test_name`, e.g. what was originally 104 is now 118, 105 is now 104, 106 is now 105, etc.

kgauthreaux:
@echeng06 hmm not sure how that happened with the CSV file, sorry!

echeng06 (changes requested):
Sorry a few more things
* `e2f-ftldivp-m-081`: Should be "If FTDAMADN = 1"
* `e2f-ftldivp-m-116`: Similarly should be "If FTDAMADN ne 1"

I think we need to relook at the Scan Model variables; apologies I missed this in my initial review and asked for the wrong thing:
* `024` and `108` make FTDSMMAM reliant on FTDSMMAN
    * If FTDSMMAN is 1-4 then FTDSMMAM cannot be blank
    * If FTDSMMAN is blank or 9 then FTDSMMAM must be blank
* `055` and `113`: Same but for FTDFDMAM, reliant on FTDFDMAN 
* `086` and `118`: Same but for FTDAMMAM, reliant on FTDAMMAN

hewanadun:
- 009 should be corrected to `If FTDSMRI = 1, FTDSMYR cannot be blank`

hewanadun (approved):

hewanadun:
@echeng06 Please approve. Thanks!

echeng06 (approved):

PR #237, opened 2024-10-30
UDS IVP - Uploading A1 requested changes

kgauthreaux:
Uploading requested changes plus updated var list for an "ADMIN" variable (will only be on A1 and A1a) and new checks associated with that

PR #238, opened 2024-10-30
UDS I4 - Upload A1 plausibility

kgauthreaux:


echeng06:
@kgauthreaux Sorry I haven't had a chance to look at this closely so I can't tell, but this seems to overlap with https://github.com/naccdata/uniform-data-set/pull/237 ; are they the same/is one more recent than the other?

To avoid confusion I think it'll be easier to just put everything in that other PR and close this one.

hewanadun:
@echeng06 Since these are I4 checks, it might be better to keep separate.

@kgauthreaux I added a new section to https://github.com/naccdata/uniform-data-set/issues/210#issue-2554408480. Please check **"Cross form error checks involving optional forms"** and make the necessary changes.

echeng06 (changes requested):
* Please rename file to `form_a1_i4_error_checks_p.csv`
* `a1-i4vp-p-1012` is a repeat of 1007 - since we want it under FRMDATEA1 I think we can just remove 1012
* `a1-i4vp-p-1017`: `test_logic` is wrong (same issue as IVP)
* `a1-i4vp-p-1019`: `short_desc` has a typo, RACERACE[UDSv3] should be RACE[UDSv3]
* Either `a1-i4vp-p-1026 - 1029` is missing `comp_vars` OR `1008 - 1009` don't need it filled out. I'm not exactly sure which convention we're going with when a variable is comparing with the V3 version of itself

hewanadun (changes requested):
- Change a1-i4vp-p-1026 to `If EDUC[UDSv3] != 99, EDUC should be greater than or equal to EDUC[UDSv3]`
- If EDUC[UDSv3] != 99 in UDSv3, are they allowed to enter 99 in I4?

kgauthreaux:
@hewanadun no I don't think they should have known the value in UDSv3 and then not know it in UDSv4. I added it as an error a1-i4vp-p-1032. It was not a problem in FVP checks because we don't ask about EDUC in the FVP

echeng06 (changes requested):
* 1032, `short_desc` should say EDUC should not equal 99, not 9

echeng06 (approved):

hewanadun:
@kgauthreaux Updated plausibility checks file has bunch of empty columns at the end. Could you please remove those. Thanks.

kgauthreaux:
@hewanadun Emily and I talked about this before, but I think it's just because I open the CSVs with Excel? I just have completely blank columns after the "questions" column and don't know how to solve the issue because if I delete a column it just gets replaced by another empty column

hewanadun (approved):

PR #240, opened 2024-10-31
LBD - Upload v3 B3L missingness & conformity

kgauthreaux:


echeng06 (changes requested):
Q&V CSV:
* `data_type` is missing for most of the variables, please update (Integer and String)

Errors CSV:
* The first two rules are under XX, please update to b3l and move rules to bottom (making them error codes 113 and 114)
* `short_desc` has the "a an integer" typo for a lot of rules, please clean up
* `b3l-lbdivp-m-008`: Should be 0-4, not 1-4

hewanadun:
@kgauthreaux No LANGXX field in LBD form headers, is that correct?

echeng06 (approved):

hewanadun (approved):

PR #241, opened 2024-10-31
LBD - Upload v3 B4L missingness & conformity

kgauthreaux:


echeng06 (changes requested):
Both in Q&V and corresponding rules:

* Everything that is 0-4 should actually be 1-4
* Everything that is 0-3 should actually be 1-3
* Some rows have "Inetgers" misspelled

Errors:
* `do_in_redcap` and `in_prev_versions` missing 
* Lots of "a an" typos in `short_desc`
* `b4l-lbdivp-m-149, 152, and 154`: The PDF says to leave this blank if not applicable, so these rules should be removed

The following is the same issue for sets Q7-Q9 but just to list them out:
* `b4l-lbdivp-m-162`: Should be reliant reliant on LBDELMED, not LBDELUS
* `b4l-lbdivp-m-164`: PDF says to leave blank if not applicable, so rule can be removed (if it's not removed it has the same issue as 162)
* `b4l-lbdivp-m-172`: Should be reliant on LBHALMED not LBHALL
* `b4l-lbdivp-m-174`: PDF says to leave blank if not applicable, so rule can be removed (if it's not removed it has the same issue as 172)
* `b4l-lbdivp-m-182`: Should be reliant on LBANXMED not LBANXIET
* `b4l-lbdivp-m-184`: PDF says to leave blank if not applicable, so rule can be removed (if it's not removed it has the same issue as 182)
* `b4l-lbdivp-m-192`: Should be reliant on LBAPAMED not LBAPATHY
* `b4l-lbdivp-m-194`: PDF says to leave blank if not applicable, so rule can be removed (if it's not removed it has the same issue as 192)

echeng06 (changes requested):
Errors:
* `b4l-lbdivp-m-001 - 002`: `full_desc` mentions FRMDATEXX, please update to FRMDATEB4L
* The 0-4 and 0-3 thing also needs to be fixed here still

image

* Sorry I missed this earlier, but for Q6-9 the PDF actually has specific instructions as shown above. So we need errors to say the a) XXXAGE fields must be 888 if the parent variable is 0 or 8 and then change the following to say "If LBDELU, LBHALL, LBANXIET, and LBAPATHY are not 1 then VAR must be blank"
    * `b4l-lbdivp-m-157` for LBDELAGE
    * `b4l-lbdivp-m-167` for LBHALAGE
    * `b4l-lbdivp-m-177` for LBANXAGE
    * `b4l-lbdivp-m-187` for LBAPAAGE
    * The med-related variables should be okay since they'll always be blank if the one parent variable is not 1, they don't care about the other 3
* So I think we should also simplify the following, else it gets messy having to account for ^:
    * `b4l-lbdivp-m-163 and 165`: Just reliant on LBDELMED
    * `b4l-lbdivp-m-173 and 175`: Just reliant on LBHALMED
    * `b4l-lbdivp-m-183 and 185`: Just reliant on LBANXMED
    * `b4l-lbdivp-m-193 and 195`: Just reliant on LBAPAMED

kgauthreaux:
@echeng06  Thanks, I saw that part of the PDF and thought that it was still covered by the parent variable being equal to 0 but I see now they still have to go in and write '888' instead of just skipping the whole Q which is very odd branching logic to have included on this form. 

For the check "If LBDELUS is in (0 or 8) then LBDELAGE must equal 888" should this actually be "If (LBDELUS is in (0 or 8) and (LBHALL=1 or LBANXIET=1 or LBAPATHY=1)), then LBDELAGE must equal 888", etc.? I just wonder if without the second part when they enter LBDELUS=0 or 8 then they will get an error that it must be blank from "If LBDELU, LBHALL, LBANXIET, and LBAPATHY are all 0 or 8 (or all not 1) then LBDELAGE must be blank" and also an error that it must be equal to '888' which would be confusing

echeng06:
@kgauthreaux Ah I see what you're saying. Yeah I think we can go ahead and write the XXXAGE rules like that, thanks!

kgauthreaux:
@echeng06 great! Is it okay if I put those in the B4 plausibility checks? since it's not really missingness but checking that it's a missing code when needed. we have been putting those in plausibility. If so, I can add it to the b4l plausibility pull request that's currently open

echeng06 (dismissed):

hewanadun (changes requested):
Check 002 should be under FRMDATEB4L

hewanadun (dismissed):

PR #242, opened 2024-10-31
LBD - Upload v3 B8L missingness & conformity

kgauthreaux:


echeng06 (approved):

hewanadun (changes requested):
Typo in 022 `short_desc` - change to SCPARATE (has PANSLTTL)

hewanadun:
@kgauthreaux One more change (sorry, didn't noticed this earlier),
In Q&V csv, the `conformity` column should be modified to `Integers .... or blank` for all variables except header vars and PACOGIMP.

kgauthreaux:
@hewanadun The questions are all required if PACOGIMP=0, and we have conditionally present/blank checks for each. I don't know why we would put "or blank" for these conditional variables since we have not been doing that unless there is an instruction saying "if unknown, leave blank" which I don't see on this form

kgauthreaux:
@hewanadun okay, I guess I just still disagree with whoever wrote it that way on RT since for both B8 and B6 we are dealing with typical conditional logic based on the first question of the form (is the participant too impaired), I mean technically for any conditional value there can be an "or blank." The only time I think it's needed is for the select all that apply Qs or if the instructions say "leave blank if unknown" although you could argue that for those they should just say "no" under missingness. But if we need this to match B6 then I uploaded an edited version

PR #248, opened 2024-11-01
Add FORMEVER to enrollment form var list

kgauthreaux:


hewanadun (changes requested):
@kgauthreaux Q&V CSV is in an outdated format, please update it to the new format and also add response labels and branching logic so we can generate the DED.

hewanadun:
@echeng06 Please generate the rule definitions for the enrollment form. Thanks!

echeng06 (changes requested):
Q&V:

* The `data_type` column is incorrect for many variables, I think this is all of them but at this point would be good to double check all:
    * ADCID and OLDADCID should be "Integer"
    * PTID should be "String"
    * FRMDATE_ENRL should be "Date"
    * INITIALSENRL should be "String"
    * ENRLGENOTH should be "Integer
    * ENRLGENOTHX should be "String"
    * GUIDAVAIL should be "Integer"
    * GUID should be "String"
    * NACCIDKNWN should be "Integer"
* ENRLTYPE `conformity` is incorrect, needs to be 1 or 2
* PTID-related conformity columns are empty - I think they should at least say "text", or maybe something "PTID-compliant text"

MC:
* FRMDATE_ENRL needs to be updated to have dates with /, not -
* `enrl-m-025`: Remove ENRLGENOTHX from `comp_vars`, also please put commas after each variable
* Need rule to say PTID must be blank if PREVENRL is 0 or 9
* Think we need a rule to ensure the NACCID format is correct, so it matches the Q&V conformity

Both:
* Based on the PDF, NACCIDKNWN looks like it should be required (missingness = Always") and `enrl-m-036` updated to only "must be 8"
    * If the conditional part _is_ intended for this variable, then `enrl-m-039` needs to be updated to say "If NACCIDKNWN is 0, 8, or blank"

Other:
* The old error checks had 4 plausibility rules - assuming they are still required can you add them here as a separate file/update as needed
    * We need to know which date to use as the release date for FRMDATE_ENRL as well

kgauthreaux:
@echeng06 Okay I think I got everything although I am not entirely sure about checks 044 and 045 since I have never had to write a 'format' check before. Also, NACCIDKNWN must be conditional due to the REDCap version - there is branching logic that hides this Q if ENRLTYPE=1. We had to put '8' as the workaround on the PDF. In the future I hope we can build on PDF first and then go to REDCap to avoid these things. Hopefully the edits I made capture the fact that it can be blank in the REDCap form

echeng06 (changes requested):
Sounds good! In that case the changes look good aside from the following nitpicks:

* Q&V CSV has an extra row, please remove

echeng06 (approved):

hewanadun (approved):

PR #249, opened 2024-11-01
UDS4 - Update A2NOT codes

kgauthreaux:
Needed to update so that the code '94' does not mean different things across UDS


echeng06 (approved):

echeng06:
@hewanadun I just went ahead and directly made the change on the draft schemas without a PR since it's trivial.

hewanadun (approved):

PR #250, opened 2024-11-02
FTLD - Upload E2F plausibility

kgauthreaux:


echeng06 (changes requested):
Both are error 1001, please update the second to 1002

echeng06:
Also in some columns for the second error there's a period after `FTDOTCHO`, please update to a comma

echeng06 (approved):
Looks good, but I wanted to also point out that the header variables are missing from the Q&V CSV (and this is referencing FRMDATEE2F). Don't know if you plan to add that in one of the existing E2F PRs or a separate one on its own. Approved otherwise!

kgauthreaux:
@echeng06 ah it does look like they were left out. I will upload an updated var list and mc checks to the open pull request I have for those

PR #204, opened 2024-09-23
Upload C2T form, variable checklist, and m/c checks

jculhane:

PR #251, opened 2024-11-02
Upload revised C2T p based on PR #204 comments

jculhane:


echeng06 (changes requested):
* Last two new errors should be under c2t, not c2
* We should also make the M/C updates made for C2
    * https://github.com/naccdata/uniform-data-set/pull/199#issuecomment-2448976757

echeng06 (approved):

hewanadun (approved):

PR #253, opened 2024-11-05
FTLD-upload-b3f-plausibility

jculhane:


echeng06 (changes requested):
* `test_name` should all say B8, I think it accidentally got auto-incremented
* Please clean up `short_desc` by removing the stuff in parentheses

echeng06 (dismissed):

hewanadun (changes requested):
@jculhane Would it be possible to merge 1004 - 1010 into one rule as follows?
If `FTDGTYP` != 8 and `GAITABN` = 1 then `FTDGTYP` should be equal to `GAITFIND` 
(Since this is FTLD module specific error, report under FTDGTYP instead of GAITABN)

echeng06:
@hewanadun I'm actually not sure we can merge them since in 1007 and 1008 they're different.

hewanadun:
> @hewanadun I'm not sure we can merge them since in 1007 and 1008 the two have different values.

Ah, I missed that. 

@jculhane Ignore my previous comment. We can still merge the cases where the values are equal if you prefer. 
- If not, please add `GAITABN = 1 ` to each check in 1004 - 1010 
(e.g. If GAITABN = 1  and GAITFIND=1 then FTDGTYP should =1)
- Do we need a check for GAITABN = 8?

jculhane:
@hewanadun Why do we need to add "If GAITABN = 1"? Isn't that implied by any non-missing GAITFIND value?

Good question about GAITABN=8. I can add a check against FTDGTYP=8, since they should select GAITABN=8 (not assessed) if it was untestable.

hewanadun:
> Why do we need to add "If GAITABN = 1"? Isn't that implied by any non-missing GAITFIND value?

That's true if the UDS packet is filled properly. We can make the system evaluate the modules only if there's an approved UDS packet. That'll make things easier. No need to add GAITABN = 1 then.

hewanadun (approved):

PR #254, opened 2024-11-05
LBD - Upload v3 B2L var list and m/c

jculhane:


echeng06 (changes requested):
Q&V CSV:
* Please update `data_type` "Num" to "Integer"
* `conformity` should also probably be updated to specify that they're integers

Errors CSV:
* `short_desc` has "a an" typo for a lot rules, please update
* ~~`do_in_redcap` and `in_previous_versions` not filled out~~

echeng06 (changes requested):
Looks like the `conformity` column got auto incremented, they should all be 9. Otherwise changes looks good!

echeng06 (approved):

hewanadun (approved):

hewanadun:
@jculhane Just noticed that `error_no` column is missing leading zeros. Please make it consistent with other forms.

PR #256, opened 2024-11-05
LBD - Upload v3 B6L var list and m/c

jculhane:


echeng06 (changes requested):
Both:
* `LBSPALRT`: According to PDF, this can be 0-10, not just 1-10, so this and `b6l-lbdivp-c-044` need to be updated

Q&V:
* Lots of "Inetgers" misspellings
* Please update `data_type` "Num" to "Integer"

Errors:
* ~~`do_in_redcap` and `in_previous_versions` not filled out~~
* `short_desc` has the "a an" typo
* `b6l-lbdivp-m-006`: The following are not necessarily present since they're dependent on other fields, so should be removed from this list:
    * 1x: LBSPYRS, LBSPMOS, LBSPINJS, LBSPINJP, LBSPCHAS, LBSPMOVE 
    * 3x: LBSPURGL, LBSPSENS, LBSPWORS
    * 6x: LBSPALRT
    * I might have missed this sort of situation in other forms - I did a quick check to see and think we might be okay since we usually check per variable, but something to keep in mind
* `b6l-lbdivp-m-041`: `short_desc` has a typo, should just be = 0

hewanadun (changes requested):
- In Q&V csv
    - All variables except header vars and `LBSPCGIM` can be blank if `LBSPCGIM` = 1, so the `conformity` should be modified to `Integers .... or blank`
    - `branching_logic` column also needs to be filled to indicate above
    
- In m/c checks
   Incorrectly filled form might lead to confusing error messages, for example, if someone mark LBSPCGIM = 1 and still filled some questions in the form,
    - System would raise error b6l-lbdivp-m-005, which is correct
    - But system could also raise any other error message 007 - 044 depending on the values entered in the form. (This happens as currently there is no variable for us to identify B6L as an optional form and ignore these checks)
    - To prevent this, we need to add `If LBSPCGIM = 0` to each missigness check. Examples,
        - 008 => If LBSPCGIM = 0 and LBSPDRM = 1, LBSPYRS cannot be blank
        - 011 => If LBSPCGIM = 0 and LBSPDRM = 1, LBSPMOS cannot be blank

hewanadun:
@echeng06 Please check and approve this PR.

echeng06 (approved):

PR #259, opened 2024-11-05
LBD - Upload v3 D1L var list and m/c

jculhane:


echeng06 (changes requested):
Both:
* Questions 4f-4l from the PDF is missing here

Q&V
* Some typos of "Inetgers" in `conformity`
* Last two variables are missing `data_type`
* `packet` should be IL, not I

Errors:
* `short_desc` "a an" typo
* ~~`do_in_redcap` and `in_prev_versions` missing for most rows~~

echeng06 (dismissed):

hewanadun (commented):
Fix `error_no` column, rest looks good.

hewanadun (approved):

PR #262, opened 2024-11-06
LBD - Upload v3 B3L plausibility

kgauthreaux:


echeng06 (changes requested):
* For all, please remove the mentions of forms (all the "UDS Form X" or "Form X") from `short_desc`
* `b3l-lbdivp-p-1016` - `b3l-lbdivp-p-1042`: I think the test logic was accidentally applied to `short_desc` and `full_desc`. "and" is supposed to be "then" and the "ne" should be "must equal"
* `b3l-lbdivp-p-1045`: Doesn't need `comp_forms` filled out
* `b3l-lbdivp-p-1006 - b3l-lbdivp-p-1010`: Not sure if we need a rule basically opposite of these saying if all of those variables are 0 then LBCMRTRM cannot be 2? If we don't on purpose then please disregard
* `b3l-lbdivp-p-1011 - b3l-lbdivp-p-1015`: Same situation as above for LBCMRIGD

kgauthreaux:
@echeng06 regarding 
- b3l-lbdivp-p-1006 - b3l-lbdivp-p-1010: Not sure if we need a rule basically opposite of these saying if all of those variables are 0 then LBCMRTRM cannot be 2? If we don't on purpose then please disregard
- b3l-lbdivp-p-1011 - b3l-lbdivp-p-1015: Same situation as above for LBCMRIGD

I do think we need a check like this but I included it on the Form D1L plausibility checks (which are not finished yet) since I could not figure out a way to write them on this form that complied with the rules!

echeng06 (changes requested):
@kgauthreaux sounds good! As for the other changes, in `b3l-lbdivp-p-1042` the `full_desc` text looks like it got kind of messed up but changes look good otherwise!

kgauthreaux:
@echeng06 oops a find & replace error. Fixed it!

echeng06 (approved):

hewanadun:
@kgauthreaux 
- b3l-lbdivp-p-1025 `short_desc` and `full_desc` text is messed up

echeng06:
@kgauthreaux the same issue is on the FVP PR so you can fix it there: https://github.com/naccdata/uniform-data-set/pull/293

PR #263, opened 2024-11-07
LBD - Upload B9L v3 FVP

jculhane:


echeng06 (changes requested):
Q&V:
* Should the packet be FL to match the errors? 

Errors MC:
* `b9l-lbdfvp-c-008`: `short_desc` should be 1-7

hewanadun (changes requested):
Similar issues as IVP
- For checks 003 - 014, it seems blank is an accepted value. Please update the Q&V CSV and m/c checks to reflect this.
- Fix check 029 => If SCCOSCVR ne 3, SCCOTH **must be** blank

@jculhane - If 777 is entered for a question, do we need to check whether the value is provided in the previous visit?

echeng06 (changes requested):
Plausibility:
* Error code `b9l-lbdfvp-p-1006` is used twice, need to change one to 1007

echeng06 (dismissed):

hewanadun:
@jculhane For 1006 & 1007 => `If VAR=777, VAR[prev_vis] should be in (15 - 110 or 777)`

hewanadun (approved):

PR #265, opened 2024-11-08
LBD - Upload v3 B4L plausibility

kgauthreaux:


echeng06 (changes requested):
* `b4l-lbdivp-p-1002 - 1005`: `short_desc` and `test_logic` should be checking for 0, I think they got auto-incremented

kgauthreaux:
@echeng06 oops yes looks like it was an auto fill error!

echeng06 (approved):

kgauthreaux:
@echeng06 okay uploading with checks for those age of onset variables to be 888 and also some extra checks against the UDSv5 NPI which measures some of the same things

echeng06 (approved):

PR #266, opened 2024-11-08
FTLD - upload b9f plausibility

jculhane:


echeng06 (changes requested):
For `b9f-ftldivp-p-1004 - 1010`:
* `test_name` should be referencing PPASYNT not PPASYN
* I might be looking at an old version of D1a but the first 2 mappings look incorrect and also all of `short_desc` says 1
    * PPASYNT = 1 (Logopenic) -> FTDCPPAS = 3 (Logopenic)
    * PPASYNT = 2 (Semantic) -> FTDCPPAS = 1 (Semantic)
    * PPASYNT = 3 (Nonfluent/agrammatic) -> FTDCPPAS = 2 (Nonfluent/agrammatic)
    * PPASYNT = 4 (Primary progressive apraxia of speech) -> No mapping
    * PPASYNT = 5 (PPA other/not otherwise specified) -> FTDCPPAS = 4 (PPA not otherwise specified)

jculhane:
Yes, that's an old version of D1a - they switched the order. I will try to review / upload D1a tomorrow so you'll have the updated version for comparison. Will wait to upload this at the same time.

echeng06 (changes requested):
Since FTDCPPA, FTDCPPAS, and FTDBVFT are conditional, do we need to update the checks to also ensure that their gate variables are also the correct value and/or allow for blankness? e.g. 

* `1003`: If PPASYN is not 1 then FTDCPPA must be 0 **or blank**
* `1013`: If FTDSYN is not 1 then FTDBVFT should be 0 **or blank**
* `1004 - 1010`: If PPASYNT= 1 **and FTDCPPA is 1** then FTDCPPAS should = 1
    * On second thought maybe not this one, since without the addition I think other rules would force FTDCPPA must be 1 (since PPASYNT/FTDCPPAS is not blank) which I think would actually be the correct behavior in this situation. But let me know what you think.

Also think we need another rule to validate this statement here: 
image

e.g. "If Q1 FTDPPASL = 1 then at least one of Q2-11 must be 2 (definitely present)"

jculhane:
I updated 1003, 1012, and 1013 to account for conditional vars - I don't think we need to update 1004-1010. I also added two rules to validate Q1 and Q14.

echeng06 (approved):

hewanadun (approved):

PR #269, opened 2024-11-10
LBD - E1L IVP plausibility and FVP var list and M/C

jculhane:


jculhane:
Correct - FVP plausibility is the same as IVP

echeng06 (dismissed):

hewanadun (changes requested):
In FVP, since the rest of the form can be left blank if `LBGNEWGN = 0`
- In Q&V, update rest of the variables to be `Conditional` and update `conformity` column to allow blank
- In m/c checks, need to add `If LBGNEWGN=1 and ....` to all `cannot be blank` checks
    e.g.  008 =>`If LBGNEWGN=1 and LBGLRRK2 = 1, then LBGLRKIS cannot be blank`
- We need a check to verify that the form is left blank if `LBGNEWGN = 0` 
    (I think vice versa is covered by current checks?)

If `LBGNEWGN=1`, do they have to pick a value for each question or can they update only some of the questions? From the error checks it seems they have to fill all, just checking whether that is the correct assumption.

jculhane:
Yes, if LBGNEWGN=1, they have to pick a value for each question

hewanadun (approved):

PR #270, opened 2024-11-10
LBD - upload c1l v3 ivp plausibility

jculhane:


echeng06 (approved):

hewanadun (changes requested):
Can we modify 
- 1002 to `If LBNPFACE is in (0-7) and LBNPNOIS is in (0-13), ...`
- 1003 to `If LBNPNOIS is in (0-13) and LBNPPARD is in (0-13), ...`

jculhane:
> Can we modify
> 
> * 1002 to `If LBNPFACE is in (0-7) and LBNPNOIS is in (0-13), ...`
> * 1003 to `If LBNPNOIS is in (0-13) and LBNPPARD is in (0-13), ...`

@hewanadun These edits aren't strictly necessary (LBNPNOIS will always be in (0-13) if LBNPFACE is in (0-7), and LBNPPARD will always be in (0-13) if LBNPNOIS is in (0-13). Happy to add if it helps you/Emily transfer them into code though - let me know.

hewanadun:
@jculhane We need to add it in the rule definition to prevent the validator code crashing if the form is not filled correctly (i.e. if LBNPNOIS left blank even when LBNPFACE in 0-7). Having it in error checks helps automation, otherwise we have to add that manually to the definition. Thanks!

hewanadun (approved):

PR #271, opened 2024-11-10
LBD - Upload v3 B6L IVP and FVP plausibility

jculhane:
I thought I had previously upload the IVP plausibility checks but I don't see them - let me know if this is a repeat/conflict.


hewanadun (changes requested):
In IVP plausibility checks,
- As explained in m/c checks PR https://github.com/naccdata/uniform-data-set/pull/256#pullrequestreview-2427720568,
   In checks 1002 to 1005 - Need to add `If LBSPCGIM = 0 and ...`

In FVP plausibility checks,
- Since error codes starting from 1006, were you planning to copy checks from IVP 1001 - 1005 here? It seems they should be included in the FVP checks too.
- For checks 1007 - 1013, 
    - Do we evaluate these checks **only if form B6L is filled in both current visit and previous visit**? If so, those checks need to be modified as follows,
       - If LBSPCGIM =0, If LBSPCGIM[prev_vis]=0 and LBSPDRM[prev_vis] = 1, then LBSPDRM should equal 1
- What about the cases where BL6 is filled in prev visit and not filled in current visit or vice versa? Are those conditions need to be checked?

jculhane:
@hewanadun We should only evaluate checks 1007-1013 if Form B6L is filled in both current visit and previous visit. I don't think I understand the modification you suggested. Why do we need to check if LBSPCGIM[prev_vis]=0? Let's say we're checking against a v3 visit that didn't have the current error checks, where someone marked LBSPCGIM = 1 and still filled some questions on the form. I think these checks are still valid as-is.

jculhane:
Gotcha. I think we can evaluate 1007-1013 even if LBSPCGIM =1 in the previous visit. 

There are cases where participants can be cognitively impaired due to something other than neurodegenerative disease (e.g., depression, medication) that is reversible, so they are allowed to mark LBSPCGIM=1 in previous visit and =0 in current visit.

echeng06 (changes requested):
* `1004` on both should be under LBSPLEGS not LBSPTRT
* `1005` on both should be under LBSPDRM not LBSPTRT

FVP:
* `1008`: Typo in `short_desc`, (0=99) should be (0-99), `full_desc` is also wrong (should say should be greater than previous visit)

echeng06 (approved):

hewanadun (approved):

PR #272, opened 2024-11-10
LBD - Upload v3 IVP E2L plausibility

jculhane:


echeng06 (changes requested):
* `LBIDATS` misspelled as `LIBDATS` in a few places
* All of the D1b variables are conditional and will be blank if BIOMARKDX (Were any biomarker results used to support the current etiological diagnosis?) is 0 (namely since it will cause IMAGINGDX to be blank). Do we need to account for this?

echeng06 (approved):

echeng06 (changes requested):
Whoops sorry approved prematurely

* `comp_forms` should say e2l, not e3f
* I think we can remove/combine 1009 and 1011 - they could have only one of LBIAPET or LBITPET. So if they're BOTH 0 then PETDX must be 0 or blank

echeng06 (approved):

PR #276, opened 2024-11-11
LBD - Upload v3.1 B2L IVP m/c

jculhane:


echeng06 (dismissed):

hewanadun (dismissed):

hewanadun (changes requested):
@jculhane @echeng06 
For the forms that the error checks are different b/w LBD v3 and v3.1 we can't use the same error codes for both versions. Need some way to differentiate.
For v3.1, can we use the format `<form>-lbd3.1ivp-<type>-<number>`?

hewanadun (approved):

PR #277, opened 2024-11-11
LBD - Upload v3.1 B6L IVP var list and m/c

jculhane:


echeng06 (changes requested):
* Error codes need to be updated to have the 3.1 in it, otherwise looks good!

hewanadun (approved):

PR #278, opened 2024-11-11
LBD - Upload v3.1 B9L IVP var list and m/c

jculhane:


echeng06 (dismissed):

hewanadun (changes requested):
- Update Q&V `conformity` column to `Integers .... or blank` to match with conformity checks
- Error codes need to be updated to differentiate v3.1, see https://github.com/naccdata/uniform-data-set/pull/276#pullrequestreview-2436233722

hewanadun (approved):

PR #279, opened 2024-11-13
LBD FVP - Upload v3 B4L missingness & conformity

kgauthreaux:


echeng06 (approved):

hewanadun (changes requested):
- Check b4l-lbdfvp-m-154 contradicting with check b4l-lbdfvp-p-1006. I think it should be changed similar to IVP check b4l-lbdivp-m-157
    - it seems other similar age variable checks are OK, but double check to confirm
- For questions 6 - 9 in PDF, since 777 is allowed value for age of onset, do we need to modify the checks for XXXMED to account for 777 in XXXAGE?  Example, 
    - 156 => `If LBDELUS = 1 and LBDELAGE != 777, then LBDELMED cannot be blank`
    - 157 => `If LBDELUS != 1 0r LBDELAGE = 777, then LBDELMED must be blank`

kgauthreaux:
@hewanadun It looks like I forgot to update the check for b4l-lbdfvp-m-154 but got it for all of the other age checks. I don't think we need to change the checks for the XXXMED variables because they should still answer the XXXMED question even if they entered age at the previous visit. E.g., they should only skip Q6b LBDELMED if LBDELUS =0 or 8 OR if LBDELAGE=888. 

I'm not sure if we need to update the check to include 888 instead or if this is take care of because they should ONLY enter 888 if LBDELUS= 0 or 8 so basically if LBDELUS=1 then LBDELAGE will not be 888 and Q6b LBDELMED must not be blank.

This part of this form makes my head hurt

kgauthreaux:
@hewanadun yes we should probably modify that one in the IVP & FVP. It can still equal 999 though so we don't need to include that

hewanadun (approved):

PR #281, opened 2024-11-13
LBD - Upload v3 B4L FVP plausibility

jculhane:


echeng06 (changes requested):
* `b4l-lbdfvp-p-1018` I think is supposed to be referencing LBDAGE, not LBANXAGE

hewanadun:
- For checks 1018 - 1022, I think we should check equality only if they enter a valid age. 
   For example 1008 => `If LBDAGE in (15 - 110) and LBDAGE[prev_vis] in (15 - 110), LBDAGE should equal LBDAGE[prev_vis]`
- If 777 is entered for a question, do we need to check whether the value is provided in the previous visit?

jculhane:
Yes, they can keep entering 777. Fixed!

echeng06 (approved):

hewanadun (approved):

PR #285, opened 2024-11-14
UDS - Upload D1a missingness & conformity

kgauthreaux:


echeng06:
@kgauthreaux is it possible to also get an updated version of the PDF uploaded (or point me to where it is otherwise)? From a different form I was told a few things in D1a got switched up, so I just want to make sure I'm looking at the right thing as I review.

kgauthreaux:
@echeng06 okay, uploaded! although D1a didn't change much, but we did get rid of MODE since the CTF said it's not relevant and confusing for a form they fill out after visiting with the participant. D1b is the one that has more significant changes

echeng06 (changes requested):
(Sorry for the billions of edits - a lot of stuff I thought was missing was actually being handled once I wrapped my head around all the branching logic so now we're just down to these :))

Q&V:
* The headers are messed up, there's two `missingness` columns 

Errors:
* `d1a-ivp-m-156`: Should be under ANXIET not MAJDEPDX, also need to fix `test_name`

Does this imply we need a rule saying at least one of them is 1 (the "select one or more as Yes")? Wasn't sure since Section 3 has similar wording but explicitly says they can all be blank. I don't know if it'd make sense to at least have an alert on this one.
image

kgauthreaux:
@echeng06 hmmm yes I think that is the intention for Q9, that they would check at least one of those. I can add a check for that. Before I upload, I could not figure out what was wrong with the test_name for d1a-ivp-m-156. Currently it says "If ANXIET =1, at least one anxiety disorder must be present" which seems right to me

echeng06:
@kgauthreaux Hm yeah taking a look again I can't tell what I was referring to either 😅 I went back and forth a lot on this one so it might just be something I accidentally left in, you can ignore that change request

kgauthreaux:
@echeng06 okay no worries! I will upload as is, if it proves to end up being an issue we can change it

echeng06 (approved):

hewanadun (approved):

PR #286, opened 2024-11-15
UDS - Upload D1a plausibility

kgauthreaux:


echeng06 (changes requested):
* `d1a-ivp-p-1008`: Reword `short_desc` to "If RMMODEC2C2T=1 and MOCBTOTS ~~should not be~~ <4 then NORMCOG should not equal 1"
* `d1a-ivp-p-1004`: `short_desc` has the very last group saying "MCICRITFUN=2", should be "MCICRITFUN=1"

kgauthreaux:
@echeng06 just a heads up, I also updated d1a-i4vp-p-1004 and d1a-i4vp-p-1005 to 'ne 1' instead of "=0" since I realized those are "1 or blank" variables so they will not be equal to 0 upon submission (just later in the Quick Access file after transforming the data)

echeng06 (approved):

PR #295, opened 2024-11-16
LBD - Upload v3.1 E2L plausibility

kgauthreaux:


echeng06 (changes requested):
Same issues as IVP: https://github.com/naccdata/uniform-data-set/pull/272

* `LBIDATS` misspelled as `LIBDATS` in a few places
* `comp_forms` should say e2l, not e3f
* We can combine 1009 and 1011 - they could have only one of LBIAPET or LBITPET. So if they're BOTH 0 then PETDX must be 0 or blank

echeng06 (approved):
Looks good! @hewanadun please also approve, thanks!

hewanadun (approved):

PR #297, opened 2024-11-17
UDS - Upload D1b var list, m/c, and plausibility

jculhane:


echeng06 (changes requested):
Q&V CSV:
* Headers are different format/missing, please update

Errors MC:
* `d1b-ivp-m-095`: This is correct, just noting it looks like it isn't ever possible for STRUCTDX to be 0 based on the branching logic
* `d1b-ivp-m-250`: Should be "If OTHCOG is blank"

Errors Plausibility:
* `d1b-ivp-p-1002` and `1003`: `short_desc`, `test_logic`, and `comp_vars` are wrong/swapped

jculhane:
Q&V CSV: Headers are correct - the mode variables were removed on this form after the pilot (see revised PDF in the Github folder)

Errors MC: d1b-ivp-m-095: I noticed this while working on these rules. It bothers me but at this point we won't change it.

Fixed other issues.

echeng06 (approved):

hewanadun (changes requested):
@jculhane 
`test_logic` and `comp_vars` incorrect in d1b-ivp-m-146, d1b-ivp-m-166, d1b-ivp-m-186

hewanadun (approved):

PR #298, opened 2024-11-18
LBD - Upload v3 D1l IVP plausibility

jculhane:


echeng06 (changes requested):
* `1001`: Should be under FRMDATED1L not FRMDATEXX
* `1002` and `d1l-lbdivp-p-1003` Please remove the "Form B3L" from `short_desc` 
* `1002`: LBUMTRFA is also listed twice in `short_desc`
* `1006`: `1004` checks RIGDNECK while `1006` doesn't, does it need to be added to 1006?
* `1036`: in `short_desc`, "If MODEB5 != 0 **and** APA = 0"
* `1066`: Comparing to Form D1a, I think this is supposed to be LBDSYNT is 2 (Parkinson’s disease). May also need another rule for "If LBDSYNT is 1 (Dementia with Lewy bodies) then LBCOGDX must be 1".

jculhane:
Fixed all. For 1066, Parkinson's is the weird case that isn't on D1b so we take the next best case from D1a. 1065 covers the LBD case.

echeng06 (approved):

hewanadun (approved):

PR #301, opened 2024-11-18
UDS - Upload A5D2 var list and m/c

jculhane:
Couple of small changes from UDS pilot feedback


echeng06 (changes requested):
* `a5d2-ivp-m-094`: Test logic incorrect, should be "IF ANGIOCP ne 1 and CAROTIDAGE is not blank"

Also just want to confirm 093 and 094 were the only two rules that were updated - it's all I see but the diff is large so I might have missed it.

jculhane:
Fixed. Correct - 093 and 094 were the only rules that were updated.

echeng06 (approved):

hewanadun (approved):

PR #302, opened 2024-11-19
LBD - Upload v3 B1L plausibility

kgauthreaux:


echeng06 (changes requested):
* `b1l-lbdivp-p-1007`: `short_desc` referencing LBSAGEFL instead of LBSAGETR
* `b1l-lbdivp-p-1010`: 777 is not valid here for IVP
* LBCMATRM misspelled as LBDMATRM in several places

echeng06 (approved):

echeng06:
@kgauthreaux Sorry could you also update these rules to match the FVP update, just so the two are consistent. 

`Ah sorry my request for changing "is non-missing (15-110 or 777)" to "is not 888 or 999" was for mainly for 1002 - 1012. It's just easier to validate on our end.`

hewanadun (changes requested):
@kgauthreaux 

- For comparisons with D1L or B9, according to the Q&V of those forms, the variables being compared can be left blank. So for checks like `If LBSAGERM is non-missing (15-110) then Form D1L LBCAREM should not equal 0`, what should be the case if LBCAREM is blank? Is it allowed or not?
- For b1l-lbdivp-p-1011, what should happen if B3 not submitted?

kgauthreaux:
@hewanadun I honestly don't think we need to do anything for either case. We are checking to see if the values disagree when they are non-missing. It's okay if the info is blank in one form due to skips but filled in another.

hewanadun:
> * `b1l-lbdivp-p-1010`: 777 is not valid here for IVP

Above is not fixed yet in `short_desc`

echeng06 (approved):

hewanadun (approved):

PR #303, opened 2024-11-19
LBD FVP - Upload v3 B1L plausibility

kgauthreaux:


echeng06 (changes requested):
* `b1l-lbdfvp-p-1007`: `short_desc` referencing LBSAGEFL instead of LBSAGETR
* LBCMATRM misspelled as LBDMATRM in several places
* `1014 - 1016`: Missing comp_vars
* `1017-1020`, `1022`, and `1024`: Could we change these conditions to be "if VAR[prev_vis] is 777 then VAR cannot be 777? I _think_ it's basically checking the same thing (lmk otherwise), but I don't think we currently support checking the previous record in the then clause like this
* Also its not currently defined what VAR should be if VAR[prev_vis] is 888 or 999, do we need a restriction on it? Maybe at minimum it cannot be 777 (and we could add that to the previous change)?

hewanadun:
@kgauthreaux @echeng06 
I think it is allowed to keep entering 777 in followup visits, right? (For example, in visit N they enter a valid value, then in N+1, N+2, N+3, ... they are allowed to enter 777)
If so, it would be easier if we separate out 777 validation from checks like `b1l-lbdfvp-p-1013` and add it to checks like `b1l-lbdfvp-p-1017`
- 1013 => If LBSAGERM[prev_vis] is in (15-110) and LBSAGERM is in (15-110), then LBSAGERM should equal LBSAGERM[prev_vis]
- 1017 => If LBSAGERM=777, LBSAGERM[prev_vis] should be in (15 - 110 or 777) 
  (This will cover the case for VAR[prev_vis] is 888 or 999, it is not allowed to enter 777 in that case)

kgauthreaux:
@echeng06 @hewanadun okay, I tried to update the checks as requested, but might have misunderstood something so sorry if it's still not right!

echeng06:
@kgauthreaux Ah sorry my request for changing "is non-missing (15-110 or 777)" to "is not 888 or 999" was for mainly for 1002 - 1012. It's just easier to validate on our end.

For 1013 - 1016, 1021, and 1023, they should look like what Chandima mentioned above:
* If LBSAGERM[prev_is] is 15-110 **and LBSAGERM is 15-110** then LBSAGERM should equal LBSAGERM[prev_is] 
* i.e. if **both** the previous and current variable is an actual age (not 777, 888, or 999) they should be the same

The updates for the 777 ones though (1017 - 1020, 1022, and 1024) look good to me though!

echeng06 (approved):

hewanadun:
@kgauthreaux Same questions as IVP https://github.com/naccdata/uniform-data-set/pull/302#pullrequestreview-2473385265

hewanadun (approved):

PR #310, opened 2024-11-22
UDS I4 - Upload D1a plausibility

kgauthreaux:


echeng06 (changes requested):
* `d1a-i4vp-p-1011`: We can just say VAR[UDSv3] instead of both VAR[UDSv3][prev_vis], but also wondering why we're comparing to DEMENTED against NORMCOG since it looks like both were also in UDSv3?

kgauthreaux:
@echeng06 Due to fluctuating cognition would be very normal/common to go from NORMCOG 0 to 1 or 1 to 0 from one visit to another as well as DEMENTED 0 to 1 or 1 to 0, since there are 2 other levels of cognition (MCI and impaired, not MCI) in between these two. So we really would just need to flag if someone had dementia at one visit and then was completely cognitively normal at the next

hewanadun (approved):

PR #321, opened 2024-12-03
LBD - Upload IVP header vars, M/C, and P

kgauthreaux:


echeng06 (changes requested):
* Error codes referencing `ftld`, need to update to `lbd`
* In M/C CSV, there is an extra top row with "Conformity/missingness checks" that needs to be removed

echeng06 (approved):
Sorry just noticed M/C and P both have an empty row at the bottom if you can clean those up, otherwise looks good!

echeng06 (approved):

PR #326, opened 2024-12-11
Update non-UTF-8 characters

echeng06:
Updates non-UTF-8 compliant characters. Also fixes the `Condtional` typo.

Done to ensure no weird characters show up when generating the DED. Also fixed the error checks since we want those updated as well when we import the error checks into REDCap.


echeng06:
Some of the LBD stuff go screwed up, please ignore this PR for now.

PR #333, opened 2024-12-18
Uploading edits to A1

kgauthreaux:
Fixing Q numbers and race/ethnicity var list


echeng06 (changes requested):
I
* 1008 and 1009: MARISTAT misspelled in full_desc

I4: 
* 1013 and 1014: MARISTAT misspelled in full_desc
* 1029: MARISTAT - should be Q11

Rest looks good otherwise, thanks!

echeng06 (approved):

PR #340, opened 2024-12-21
Editing B5 plausibility checks

kgauthreaux:
From Teams:
1001 - 1020 need to account for blank B9 variables (maybe just on the "should equal 0" ones, your call) 1026 and 1028 - short_desc should be "not 1" instead of "0 or 2" There was a discussion on whether we need to account for variables being 9 (Unknown) - I do think we might want to account for it because right now everything will raise an error/alert if it's set to 9 I think all of the above could be resolved by changing all the even codes to basically be the opposite of its corresponding case (the odd error codes), e.g. 1002 - 1020: "should not be 1" (so will pass if it's 0, 9, or blank) 1022 and 1024: "should not be 1" (so will pass if its 0, 2, or 9) 1026 and 1028: "should not be 1" (so will pass if its blank) If 9 is also allowed in the opposite case, all the odd error codes need to be updated to include 9 as well


echeng06 (changes requested):
* 1021: `test_logic` ANXIETY should be "ne 1" instead of "in (0,2)" since it can be 9
* 1023: `test_logic` same as above for MAJORDEP and OTHERDEP
* 1024: `short_desc` "should equal not equal" typo
* 1025: `test_logic` ANXIET should be "ne 1" instead of "in (0, 2)" since it can only be 1 or blank
* 1028: `full_desc` should be "and" not "or"

echeng06 (approved):

PR #345, opened 2024-12-27
Bugfix/typos

echeng06:


hewanadun (dismissed):

kgauthreaux:
@echeng06 see my responses to your teams message below. All other edits here look good to me!

- A2 and B9 have duplicate errors, namely "a2-ivp-p-1009 - 1011" and "b9-ivp-p-1001 - 1003" are the same. I have no preference on which one to remove it from. **KG: Let's leave these in B9 since it's required. I can submit a pull request to remove them from A2**
- B9 "b9-ivp-p-1006 - 1007" (for both I/I4) has inconsistent short_desc vs full_desc/test_logic - they're both technically correct but since it is coded based on short_desc the error reported by full_desc is confusing. We should swap the order in full_desc/test_logic so it's consistent with short_desc, or vice versa. **KG: I think what you had in commit [b47dcec](https://github.com/naccdata/uniform-data-set/pull/345/commits/b47dcec6c16b64528383dbbd433027a1392f7650) looks fine!**

PR #350, opened 2025-01-14
Add PTID character limit to enrl form

kgauthreaux:


hewanadun (approved):

hewanadun (changes requested):
Actually, let's add the check to `OLDPTID` instead of `PTIDCONF`. 
PTIDCONF already has a similarity check to PTID, so it would be caught if they are different.

hewanadun (approved):

hewanadun:
@kgauthreaux Sorry for not noticing this earlier, PTID length can be <= 10, doesn't need to be exactly 10. Could you please correct the checks.

PR #352, opened 2025-01-16
Kgaut add ptid check to enrl

kgauthreaux:


hewanadun (changes requested):
@kgauthreaux We already have GUIDs longer than 10 characters in the system. NACCID lookup is only checking for it to be alpha-numeric string (no limit). Can you relax the error check to match that?

kgauthreaux:
@hewanadun okay, you might have to check with Laura on the REDCap project since I think we got that limit from the REDCap form which in turn I think got it from the NIA GUID generator site. shouldn't I just delete the check then? It is essentially just any text and we don't write checks for that

hewanadun (approved):

PR #353, opened 2025-01-16
Kgaut upload module fvp var lists

kgauthreaux:
For the FVP DEDs!


echeng06 (approved):
Looks good to me, thanks! I also fixed the encoding and updated some weird characters, so if those look good to you we can go ahead and merge.

PR #358, opened 2025-01-29
Fix LBD and FTLD plausibility files

kgauthreaux:


echeng06 (changes requested):
* For B3L, should be **and** PDNORMAL != 1

hewanadun (changes requested):
- form_b3l_fvp_error_checks_p.csv error codes got replaced to `ivp` => change back to `fvp`
- If there are no changes to UDS form B3 checks, please download the version in the main branch and replace the files in this branch with that version.

kgauthreaux:
@hewanadun I uploaded the B3L fvp file with that fix. I don't know how to do the other thing you mentioned so if you or @echeng06 can do that, that would be great

echeng06:
@kgauthreaux @hewanadun I can handle it, since it looks like I also need to fix some merge conflicts with the formatting fixes I added the other day.

hewanadun (approved):

echeng06 (approved):

PR #359, opened 2025-01-29
Edit B9 COGAGE check

kgauthreaux:


echeng06 (changes requested):
I think we also want to update `b9-ivp-m-057` for consistency - in some non-REDCap case they could answer all "No/Unknown" but still give something for COGAGE, and the question is specifically worded to require those: 

"If any of the cognitive symptoms in 9a-9h are present, at what age did they begin?"

echeng06 (approved):

PR #364, opened 2025-02-06
A1 - Remove '9999' code from BIRTHYR var list

kgauthreaux:
Uploading this now as I hope we can officially  approve on Monday, but since I know it's incorrect I wanted to just go ahead and fix it to be ready. You can wait to review once it's approved.


kgauthreaux:
@echeng06 @hewanadun okay, I think this is okay to move forward now that we reviewed this AM!

hewanadun (approved):
Sounds good!

@kgauthreaux this will require re-generation of DED, can you add a column to RT tracker to track that task (if it is not there already), so we don't forget it at release.

PR #365, opened 2025-02-11
Form B6 GDS algorithm

kgauthreaux:
@echeng06 @hewanadun here is my attempt at coding an error check for the GDS algorithm on pg. 52 of the [UDSv4 coding guidebook](https://files.alz.washington.edu/documentation/uds4-ivp-coding-guidebook.pdf):

Calculate the sum of values for all checked “Yes” or “No” answers and enter the total score in the space provided. The calculation
may include a maximum of three unanswered items, and the final sum must be prorated for the number of unanswered items (see
instructions below for prorating scores). If more than three items are unanswered, however, the test must be considered incomplete and the Total GDS Score entered as 88=Did not complete.

PRORATING SCORES (what to do if the participant does not answer up to three items): If up to three of the 15 items are
unanswered (i.e., responses are 9=Did not answer), add the total score on the completed items plus an estimated score for the
unanswered items to get a total score. The estimated score for unanswered items is calculated as:

 **(Total score of completed items / # of completed items) * (# of unanswered items)**

You may get a fractional answer that requires rounding. For example, if the subject got a score of 5 for 12 completed items, then the estimated total score is **5 + [(5/12) * 3] = 6.25**. Since the decimal portion of this value is <0.50, the total GDS score is 6.


hewanadun (changes requested):
@kgauthreaux partial score calculation formula looks good. Can we use the error code 1004 for that check (I think it was originally 1004 and was removed).
Also, 1003 descriptions need to updated (since now it is applicable only if all fields are valid)

hewanadun:
@kgauthreaux bumping this up for visibility

hewanadun (approved):
I removed 1006.

PR #371, opened 2025-02-27
Add plausibility checks for EXPNOTAPP and EXPNOANS

kgauthreaux:
Adding these to follow same logic as the gender checks on Form A1. Slightly different than the REDCap @NONEOFTHEABOVE function described [here](https://app.asana.com/0/1209173861012103/1209313517314029) but gets at a similar thing - these two should not be selected if other options are selected.


echeng06 (changes requested):
`test_logic` has unrelated variables in it, also please remove extra row/columns thanks!

echeng06 (approved):

PR #357, opened 2025-01-28
Upload BDS var list and m/c (no plausibility)

jculhane:

PR #372, opened 2025-02-28
Upload Milestones error checks

kgauthreaux:
In addition to the Milestones [form](https://files.alz.washington.edu/documentation/uds3-milestone-form.pdf) and [DED](https://files.alz.washington.edu/documentation/uds3-milestone-ded.pdf), I used this [Milestones form flowchart](https://files.alz.washington.edu/documentation/milestones-flowchart.pdf) for logic which Janene recommended. One other thing I did confirm with Janene in Box A is that if Q2a is not blank, then Q3 must be blank, and vice versa, if Q3 is not blank, then Q2a must be blank


echeng06 (changes requested):
Q&V:
* PTID should update to say "Alpha numeric code up to 10 digits" instead of just "Any text"
* For PACKET should the response label just be "M"?

Error checks M/C:
* The following should be "and" not "or":
    * `milestonesv3-m-011`
    * `milestonesv3-m-014`
    * `milestonesv3-m-017`
* For all the "If PROTOCOL not in (1-3), it's cleaner to say "If PROTOCOL is blank", same for "is in (1-3) = not blank"
    * Test logic is also wrong for some of them (only checks "PROTOCOL=1", should change to "PROTOCOL is blank")

**EDIT:** Removed my previous comment about the weird dual reliance between PROTOCOL/FTLDDISC and DECEASED/DISCONT, I'm seeing now the DED has it set up that way. It's kinda weird IMO but might be better to keep it consistent, my main concern is the user will basically get double the error checks for the same issue.

kgauthreaux:
@echeng06 thanks, I will look into these. Quick Q before I upload the NP checks, is the "formver" column that is in the BDS questions and vars and M&C list okay to keep? Because if so, I may add that back into the milestones checks and keep them in the NP checks so that all of these have the same format

kgauthreaux:
@echeng06 checking in on my question above! Jessica said you had her change the "formver" column to "packet"  on BDS but that it was okay that the value for the "packet" column was '1'?

echeng06:
@kgauthreaux Oops sorry missed this. I think she's referring to this comment: https://github.com/naccdata/uniform-data-set/pull/357#issuecomment-2698632654

which @hewanadun can weigh in better on, but I think it's probably fine?

echeng06 (changes requested):
Okay sounds good!

* Can you rename the files `form_milestones_x.csv` to match the others?
* Plausibility error code/form name still has the V3 in milestones
* M/C file had a lot of empty columns/rows added, please remove
* milestones-c-008 is handled by preprocessing checks so can remove
* milestones-m-045 PROTOCOL is repeated twice in short_desc

hewanadun:
> @echeng06 checking in on my question above! Jessica said you had her change the "formver" column to "packet" on BDS but that it was okay that the value for the "packet" column was '1'?

@kgauthreaux @jculhane  @echeng06 Sorry I missed this question. I just asked to remove the "formver" column since it wasn't included for other modules/forms. "packet" column is optional, you can include it if there's a packet code for the module/form, else no need to include a "packet" column (similar to enrollment form).

kgauthreaux:
@echeng06 should we edit the pre-processing checks to specify milestones, bds, np for the PTID check or is that check basically just going to apply to all forms with PTIDs going forward without having to edit it?

echeng06:
@kgauthreaux We do need to add it there but I don't think it's the only one that needs to be updated so that can be a separate PR from TT, I think there's a few related configs on our side that we also have to build for these anyways.

echeng06 (approved):

hewanadun (changes requested):
@kgauthreaux Please move check milestones-m-047 under PROTOCOL instead of RECOGIM
Rest looks good!

hewanadun (approved):

echeng06 (approved):

PR #373, opened 2025-03-04
Update c2c2t missing codes for consistency

kgauthreaux:
Based on this task: https://app.asana.com/0/1209173861012103/1209458627716278


hewanadun (commented):
@kgauthreaux  In C2, there are spaces added in DIGFORCT and REYDTI response_labels. Is that intentional? Just checking since response_labels column for other variables don't have spaces b/w the responses.

Looks good otherwise.

kgauthreaux:
@hewanadun hm that's odd! I don't think it really matters, but just for consistency I edited it!

hewanadun (approved):

PR #374, opened 2025-03-12
Update PTID conformity across modules

kgauthreaux:


hewanadun:
@kgauthreaux We need to keep the alphanumeric restriction for GUID. "GUID definition is from NIH. alphanumeric with underscores"

kgauthreaux:
@hewanadun I thought we had to change the PTID description because alpha numeric does not include underscores? Should I add "with underscores" to the GUID description?

hewanadun:
> @hewanadun I thought we had to change the PTID description because alpha numeric does not include underscores? Should I add "with underscores" to the GUID description?

Correct, alphanumeric does not include underscore. So GUID should be "alphanumeric with underscores"

hewanadun (approved):
I made a minor change to GUID description.

PR #392, opened 2025-04-04
Upload NPv11 error checks

kgauthreaux:


echeng06 (changes requested):
Q&V
* PACKET and FORMVER required (need Missingness set to Always)
* Conditional misspelled
* NPPMIH: conformity should be 00.0 - 98.9, 99.9 
* NPPATHO: conformity only 0 and 1 (no 8/9)
* NPPATH missingness should be Always

M/C
* np-c-008 handled by preprocessing checks so not explicitly needed
* np-c-024: NPPMIH again, should be 0-98.9 OR 99.9 
* np-c-074: test_logic is wrong (needs to exclude 6 and 7)
* np-m-093: Should be "and" not "or"
* np-c-215 NPPATHO again can only be 0 or 1

Also please update the filenames to form_np_*.csv

echeng06 (dismissed):

hewanadun (changes requested):
- Variable `FORMVER` appears twice in Q&V (0b. Form version number and 21. Form version)
- Want to double check whether `NPID` is required, this is Null in some existing NP records in the database. Also, we need to remove the alphanumeric restriction in conformity check, there are hyphens, etc. in existing values.
- Do we need to add `INITIALS`? Just checking since there is `Completed by` field in PDF (but can't find it in DED)

kgauthreaux:
@hewanadun I thought I had asked Zach about NPID but maybe it was just about the conformity (Teams deleted our chat about it). Since there are blanks I removed that check. I could not find INITIALS in the NPC table so I left it off since it also was not in the DED

hewanadun:
@kgauthreaux I updated the FORMVER error codes to 003 and 004 since rule definition already uses those. Just want to double check whether you want to keep Q.21 as the question number for FORMVER, I feel Q.0b would be more appropriate as this is a header variable?

hewanadun:
@kgauthreaux bumping up for visibility, please check my question above. Thanks!

kgauthreaux:
@hewanadun yes, I was not sure if we should keep it as 21 because that's the way it is in the DED? Idk this form's DED is very weird though, e.g., I am not sure why PTID had to be an actual Q1 on the PDF instead of a header var. But since Q21 is not actually on the PDF anywhere it's probably okay to change it to Q0b as a header

echeng06 (approved):

PR #397, opened 2025-04-07
Add files via upload - B1a vars and m/c

jculhane:


echeng06 (changes requested):
* Is the Q&V CSV supposed to be named `ivp_questions_and_vars`? I'm assuming they're the same, but the packet listed in the file is I
* PACKET and FORMVER seem to be referencing the Milestones, there's also a mismatch in the M/C file (description says one thing and test logic says another), so I'm not exactly clear what it's supposed to be
* Is packet supposed to be in the error codes? Either way we also probably want everything in lowercase for consistency

The rest look okay, I might have missed a conversation on how we're handling the versions/packets/codes that address the above.

jculhane:
All of these should now be fixed. You're correct that the I and F packets are the same for this form.

echeng06 (approved):
Can we rename the files to b1a instead of b1? Looks good otherwise!

jculhane:
Yes, sorry!

echeng06 (dismissed):

hewanadun (changes requested):
@jculhane I think we need to add `VISITNUM` too.

jculhane:
Correct! Added.

hewanadun (approved):

jculhane:
@echeng06 do I need your review or can I go ahead and merge? (Not trying to rush you)

echeng06 (approved):
Sorry, approved!

PR #408, opened 2025-05-16
Add files via upload - DS C1D vars, m/c, and p

jculhane:
FVP is identical to IVP


echeng06 (changes requested):
Q&V:
* Nitpick, but could we remove the comma after the e.g. in "Conditional (e.g., skip pattern) - this just aligns with how this string is defined in previous forms (so should be "Conditional (e.g. skip pattern)", especially if now at this point people are hardcoding those strings (like me lol) 

M/C:
* Missing error checks for FRMDATE_C1D
* c1d-dsivp-c-041: Should be "or 88.8", not "ne 88.8"
* c1d-dsivp-c-053: Should be "or 88.8", not "ne 88.8"

P:
* For all, need to wrap in condition that all fields being added are not None

echeng06 (changes requested):
Plausibility:

* 1001-1010 all still need to check that the variables are not blank (not just 8/etc.)
* c1d-dsivp-p-1006 in particular needs to check that the totals are valid

jculhane:
Instead of checking for blanks, I checked the gate question - hopefully this works the same on your end.

echeng06 (changes requested):
Okay that works! One more thing, in the Q&V, DSLGNAM response label is incorrect (should be 88 not 8)

echeng06 (approved):

PR #409, opened 2025-05-16
Upload DS module B1D IVP vars & MC checks

kgauthreaux:
Deleting old REDCap files that Ben added since RIT is working on a current version of the project


echeng06 (changes requested):
Both files are improperly formatted, please update! (has extra header stuff)

Q&V
* DSONSTYR missing data_type
* DSREPMO missing data_type (kind of weird to me that it's a string but if that's what the PDF denotes then that's fine)

M/C:
* b1d-dsivp-m-076 and 077 are duplicates, I think can just remove 077
* In general for 16x I think it would make more sense for these rules to rely on whether DSDXHS_MCI and/or DSDXHS_DEM are blank, since there is nothing preventing DSDXHS_NO from being blank
    * 068, 073, 076, 080
    * e.g. for 76, it'd be "If DSDXHS_MCI **and** DSDXHS_DEM are blank, then DSONTYR must be blank"
* b1d-dsivp-m-146 - b1d-dsivp-m-185: the missingness/conformity denotation becomes incorrect (does missingness twice and then the pattern is off, probably copy and paste error), need error_code and check_type columns
* b1d-dsivp-m-252: Small typo in short_desc, should be "must **be** blank"

kgauthreaux:
@echeng06 sorry about the header stuff still being there! and yes for DSREPMO it's just a text field in the PDF and the DED just says "Numeric free-text" so I worry that if we limit it to integers we will get some errors from people over the years who may have put "Two" instead of "2"

echeng06 (approved):

PR #410, opened 2025-05-19
Upload DS Module D1D var list, m/c, and p

jculhane:
FVP differs from IVP, and v4 plausibility differs from v3 plausibility


echeng06 (changes requested):
* `forms_d1d_ivp_error_checks_p.csv` has C1D contents - I'm guessing this file can just be deleted
* FVP plausibility files need their codes updated to "fvp" instead of "ivp" and packets updated to "FDS" instead of "IDS"
* Please use `[prev_vis]` instead of `[prev_visit]` for FVP errors (just for consistency)
* 1004 in both FVP plausibility files listed under DEMENTED, should be DSCOGST (namely var_name and test_name columns)

PR #413, opened 2025-05-28
Add files via upload - A1D var list, m/c, and p

jculhane:


echeng06 (changes requested):
IVP PDF (not sure if change requests are applicable here but just in case)
* Q5, "Present - resolved" doesn't have a number label

IVP M/C
* 138, 140, 145, 147: Missing the "If" in `short_desc`

FVP Q&V
* DSSYNC and DSPNEUM have incorrect conformity (looks like auto-increment error)

FVP M/C
* 001 and 002 need to be updated to FVP (error code and packet)
* 138, 140, 145, 147: Missing the "If" in `short_desc`

FVP P:
* Please use `[prev_vis]` instead of `[prev_visit]`

jculhane:
Thanks for pointing out the PDF error - it's in the queue. All others should be fixed in the upload.

echeng06 (approved):

PR #415, opened 2025-06-05
Edit B9 plausibility checks against B8

kgauthreaux:
See description here: https://app.asana.com/1/941506444463212/project/1209173861012103/task/1210357096036245?focus=true. On B9, they can say 'no' to the gate question so the rest of the questions in the section are blank


echeng06 (approved):
Looks good!

PR #417, opened 2025-06-09
Upload D1C variable list

kgauthreaux:
Uploading variable list and current draft form for RIT to review the REDCap form. This form still needs to be piloted (currently scheduled for July 2025)


kgauthreaux:
@jculhane should be the correct CSV now

PR #432, opened 2025-07-14
DS Module - Upload B1D IVP plausibility checks for UDSv3 and UDSv4

kgauthreaux:


echeng06 (changes requested):
V3
* `1053` and `1055`: `short_desc` worded weirdly

I'm still working/testing through this but noting the above now so I don't lose track of it.

kgauthreaux:
@echeng06 honestly a miracle!

echeng06 (approved):

PR #440, opened 2025-08-08
Upload UDSv4 A1 FVP M/C and Plausibility

kgauthreaux:


echeng06 (changes requested):
M/C

* Can we change `ivp_v4` to `udsv4_ivp`? Sorry to change it up, I'm just seeing in the past we've used `udsv3` so want to make things a bit more consistent
* `a1-fvp-c-002`: `short_desc` has a weird format, can probably just change it to "FRMDATEA1 must be a date" like other forms
    * Just noticed this is in IVP too (`a1-ivp-c-147`) 
* `a1-fvp-m-024` `test_logic` needs to say "blank" not "0"
    * Also `a1-ivp-m-073` in IVP
* `a1-fvp-c-026`: `full_descr` needs to say "blank or 1" not "0 or 1"
* `a1-fvp-c-040` missing `error_type` column
* `a1-fvp-m-056`: Should be conformity rule, not missingness
* `a1-fvp-m-057`: Should be conformity rule, not missingness

P
* `a1-fvp-p-1005`: ~~Should also restrict it to not being 9.~~ Can we also use `prev_vis` in the `short_desc`?
* `a1-fvp-p-1011`: ~~Should also restrict to not being 9~~
    * As discussed over Teams we don't necessarily want to put a restriction on 9, so crossing this out but leaving it in this comment just in case we need to come back to it!

Q&V
* Conditional is misspelled

echeng06 (approved):

PR #441, opened 2025-08-08
Upload UDSv4 A3 FVP M/C and Plausibility

kgauthreaux:


kgauthreaux:
@echeng06 sounds good! and probably because looking at these error checks is just a headache lol

hewanadun:
@kgauthreaux @echeng06 

For checks like to a3-fvp-m-024, I think it might be clear if we split those into a MC check and a plausibility check. WDYT?
- MC check: If MOMETPR is in (01-12) then MOMETSEC must be present
- P check: If MOMETPR =66 and MOMETSEC[prev_vis] is not blank, then MOMETSEC cannot be blank

For 1038-1045, I'm OK with the suggested format.

echeng06 (changes requested):
Q&V:
* KIDS missing the 66 option
* SIBAGO and KIDAGO variables missing the 888 option - also issue in IVP
    * For IVP added to [Asana here](https://app.asana.com/1/941506444463212/project/1209173861012103/task/1211524915122920); please fix in a separate PR from this one

MC:

* Super nitpicky, but there are a lot of rows where the there is no space between the number and "and" in the descriptions/test logic, for example in a3-fvp-m-056 "**>0and** <=20"

P:
* 1054-1056 should be corresponding SIB2 variable, not 1
* For SIB variables, `full_descr`, should be "ETPR and OTHER_VAR", not "ETPR and ETPR"
* For KID variables, remove space between var and [prev_vis] in `short_descr`
* Missing SIBXYOB checks against prev_vis
* All the AGO <= AGD variables should also check "not blank"
    * Realizing that in IVP, AGO <= AGD variables should also check both are not blank, 888, or 999; the implementation of these checks are actually already doing this, need to update wording to match
        * Added to [Asana here](https://app.asana.com/1/941506444463212/project/1209173861012103/task/1211524915122926) 
    * Should word it like "If SIB2AGO is 0-110 and SIB2AGD is 0-110, then SIB2AGO must be <= SIB2AGD" so it just captures all and works for FVP and IVP; this is how its actually implemented in IVP currently as well
    * Includes MOM/DAD, and all SIB/KID variables

echeng06:
@kgauthreaux Ah okay good catch, in that case we actually need to update the IVP M/C rules to not allow 888. I will update the Asana issue (if you do fix it for this release please do it in a separate PR).

echeng06:
@kgauthreaux It looks like you also updated the AGD variables, which I think can be 888?

echeng06 (changes requested):
MC:
* It looks like you also updated the AGD variables, which I think can be 888?

P:
* So KIDS have a YOB rule "If (KIDS =66 and KID3YOB[prev_vis] is not blank), then KID3YOB must be present"; does SIB need these as well? That or could potentially be merged with a3-fvp-p-1265 - 1279, and same thing done to the SIBXYOB variables?

kgauthreaux:
@echeng06 hmm, I re-uploaded here: https://github.com/naccdata/uniform-data-set/pull/441/commits/579e6d9a61d32feaa3576b958d2bc48ce969097e Can you see them now?

hewanadun:
> @hewanadun that's true, though they really should not put '66' in that case...I'm trying to decide how much we care 😂 what if we just checked it for the SIB/KID YOB variables? it would save a billion additional checks but would at least trigger a sign that they should go back and edit the '66' to the new number if there is new info about additional siblings/kids

@kgauthreaux yes, checking the SIB/KID YOB variables should be sufficient.

echeng06 (approved):

PR #442, opened 2025-08-08
Upload B3 FVP plausibility

jculhane:
no change to m/c from IVP


echeng06 (changes requested):
Assuming we move `b3-fvp-p-1003` to be under GAIT, I'm not seeing any difference between this and the IVP version (I4 is different) aside from more ordered error codes. Am I missing something? Otherwise I'm not sure we need to have separate error checks?

PR #444, opened 2025-08-08
Upload UDSv4 B7 FVP Plausibility

kgauthreaux:


echeng06 (changes requested):
P
* `b7-fvp-p-1001` and `b7-fvp-p-1002`: `full_descr` should be in "if-then" form
    * IVP should be updated as well (maybe in a later release)

Q&V
* Conditional misspelled

PR #446, opened 2025-08-08
Upload UDSv4 B4 FVP Plausibility

kgauthreaux:


echeng06 (changes requested):
Looks good, but could we redo the error codes so that they're in order? e.g. make the current `b4-fvp-p-1018` `b4-fvp-p-1006`, so it aligns with the other packets (namely I4).

Also Conditional is misspelled in the Q&V file.

kgauthreaux:
@echeng06 ahhh sorry, I always forget the CSV doesn't automatically update the code with a formula like our SharePoint files do!

echeng06 (approved):

PR #448, opened 2025-08-11
Upload B1 FVP plausbility

jculhane:
No changes to M/C from IVP


echeng06 (commented):
P
* Should add Q numbers to `full_descr`
    * Probably also want to update I and I4 too
* 1005 - 1008 should also check that the previous visit was not 888 (unknown) (i.e. like 1003/1004)

echeng06 (approved):

PR #450, opened 2025-08-11
Upload UDSv4 D1b FVP Plausibility

kgauthreaux:


echeng06 (changes requested):
Looks good! I still have to finalize the rules for these so please hold off on merging until I approve again!

echeng06:
@kgauthreaux Sorry I just saw this issue on the M/C file

* `d1b-ivp-m-147`, `d1b-ivp-m-148`
* `d1b-ivp-m-167`, `d1b-ivp-m-168`

The above I think was changed to be "if OTHBIOMX is ... then". The `full_descr` is correct but the other columns (`short_desc`, `test_logic`, and `comp_vars`) are incorrect.

This only affects the DED (error checks were implemented correctly); I've added a ticket to Asana so we can add it to the next releaese! https://app.asana.com/1/941506444463212/project/1209173861012103/task/1211153804454066

echeng06 (approved):
Okay, aside from the M/C typos we need to do (can be in different PR/for next release) looks good!

PR #452, opened 2025-08-15
Upload UDSv4 A4a FVP M/C and Plausibility

kgauthreaux:


echeng06 (changes requested):
MC:
* `a4a-fvp-c-002`: `short_desc` has a typo in the date (the `dd-dd`)
* `a4a-fvp-m-015` and `a4a-fvp-c-017` are redundant
    * I think 017 is just supposed to be the conformity rule ("NEWTREAT must be 0, 1, or 9) not a missingness conditional
* Anywhere that says "If TRTBIOMARK in (0,9) **and** NEWTREAT in (0,9)" should actually be an "or" (`a4a-fvp-m-019` has the correct short_desc but the rest of the columns specify "and", rest of errors are wrong across all columns)
    * Alternatively I think we can just say "If NEWTREAT = 1 then..."? A few rules use this version. The former "guards more" if the user like sets TRTBIOMARK to 0 but NEWTREAT to 1, but the latter is simpler. Slight preference towards the latter but ultimately it doesn't matter much to me.
* `a4a-fvp-c-224` and `a4a-fvp-c-226`: Same redundancy issue
    * Additionally, 224 should also check if ADVEVENT is blank
* Same gate situation with ADVEVENT and NEWADEVENT (can be on of the two are 0,9 vs just if NEWADEVENT = 1)

P:
* `a4a-fvp-p-1002`: `full_descr`, should be Q3 not Q1

echeng06:
> @echeng06 do you mind reviewing this one? only because the IVP edit I uploaded is tied to this form change: https://app.asana.com/1/941506444463212/project/1209173861012103/task/1211186908403172?focus=true. However, if you don't have time, it's super minor and totally can wait for the next form change release!

Could we put the IVP change on a different PR so we can approve/merge them separately? I still need to write validation tests for the FVP rules and that might take me a bit.

echeng06 (changes requested):
MC:
* FVP MC still has the dd-dd issue for `a4a-fvp-c-002` in `short_descr`
* `a4a-fvp-c-012`: RMMODEA4A can only be 1 or 2
* `a4a-fvp-m-160`: Says ALZNET6, I think supposed to be STARTMO6?

echeng06 (approved):
Looks good, thanks for all those changes!!

kgauthreaux:
@echeng06 thank YOU for your patience on this one!! Sorry there were so many issues!

PR #475, opened 2025-10-14
Upload A1 informative missingness and derived vars

jculhane:


echeng06 (dismissed):
Missingness
* NACCLANGX - we can leave it here, but I feel like this is more part of the derived variable logic, not missingness (they kind of go hand-in-hand but still)

echeng06 (commented):
@jculhane I realized A1 is the form where there's a bunch of variables only collected in IVP and not FVP; they're not exactly missingness values but need to be "pulled through" in the same way.

Would it be too difficult to add/denote them to the missingness file here (I think it largely overlaps with what's already here but I ran into a couple that are missing)? Or is there a list I can reference (if it's just all the variables in IVP but not FVP I can probably generate that myself). And are there any forms other than A1 where we need to worry about that?

jculhane:
I removed NACCLANX from the informative missingness - that was an error.

I added "carry forward" logic to the TT code on the derived variables, and also added a list of all non-derived carry forward variables (including v3) to the informative missingness.

I believe A1 is the only form that has this type of carry forward variable.

echeng06 (approved):

PR #478, opened 2025-10-14
Upload B9 IM and derived variables

jculhane:


echeng06 (changes requested):
Missingness
* The variables in the first 5 rows are also in V3 and earlier - is it the same logic for those? And if so `Version` column should be "1-4" too
    * Except ALCUSE, SEDUSE, OPIATEUSE, COCAINEUSE, CANNABUSE, OTHSUBUSE 
* DECCLCOG misspelled in several places
* Several rows under `Version` say "4-Jan", guessing this is supposed to be "1-4" 
* COGMODEX has a "If is" typo under `TT code` column
* COGOTHRX is included in the DECCLCOG group but shouldn't be
* COGORI misspelled as COROGI
* ~~For missingness against prev visit:~~
    * ~~These variables may not necessarily exist in the previous visit, or previous visit could also be 777. Should it be "value of prev visit that is not empty or 777, or -4 if not ever set"? (even if that situation is technically not possible with error checks, ideally the logic here is isolated from error check assumptions).~~

~~I see a variable can have multiple conditions. I know it's tedious but I think it'd be clearer if we only have one row per variable, because we also need to consider the order of priority else they may clobber each other (even if error checks are SUPPOSED to account for it, again this logic should be isolated from those assumptions as much as possible).~~

~~For example, FRSTCHG: should first check if DECCLIN = 0 and then set FRSTCHG to 8, even if FRSTCHG happens to be 777. Then if that is not satisfied, if FRSTCHG = 777, then FRSTCHG should equal prev visit. Then if that is not satisfied, default missingness (-4).~~

echeng06 (changes requested):
Disregard! Still need the mentioned changes above though.

echeng06 (approved):

PR #479, opened 2025-10-14
Upload C2/C2T IM and derived variables

jculhane:
C1, C2, and C2T included


echeng06 (commented):
Missingness:

* In V4, all the INT variables can be 0-99, so the 88 and 95-98 will contradict/overlap
    * VERBALTEST gated rules should set to 888 instead? (some such as REYBINT do this already)
    * REYXREC gated rules also need to be modified
        * REY6REC and REY6INT should rely on REYBREC, the rule for REY6REC is also repeated twice
* "If VERBALTEST = 1 then REYFPOS should = 88" - should be "If VERBALTEST = 2"
* Doesn't really matter but I don't think MINTTOTS can ever be 88
* Assuming NACCMOCA is -4 if C2T was submitted instead?

echeng06:
One more question that popped up with testing:

* For the TRAIL* variables, e.g. TRAILARR, it looks like the legacy code subtracted 900 to make these 2 digits (e.g. 95 instead of 995) - should that also be the behavior here or is the 3 digits now intentional and should replace that?

echeng06 (approved):

PR #517, opened 2026-01-30
Adding Chinese language options to A1, A1a, A2, and A5/D2

kgauthreaux:
https://app.asana.com/1/941506444463212/project/1209173861012103/task/1212710376679326?focus=true

Will also making changes in REDCap PID 145 and let Laura know in case she wants to review


kgauthreaux:
@hewanadun looks like we never actually merged these changes which is nice! I will just close this pull request

PR #520, opened 2026-02-03
FTLD edits

jculhane:
Edit FTDRATIO var list to allow 88.8


hewanadun (commented):
@jculhane Changes looks good. Could you please remove the extra blank lines added at the end of the files, these cause issues when importing error checks to REDCap.

hewanadun (changes requested):
Just noticed that short_desc was not updated in C5F m/c checks

hewanadun (approved):

PR #532, opened 2026-04-06
Increase A3 and B9 max ages from 110 to 120

jculhane:
https://app.asana.com/1/941506444463212/project/1209173861012103/task/1213804797926971?focus=true


hewanadun (changes requested):
@jculhane A3 Q&V filename is missing the last "_" (in questions_and vars), so it is shown as a new file and hard to track the changes. Please re-upload it with the correct name and delete the incorrect file.

Also, want to double check that the updates are made only to the variables mentioned here?
https://app.asana.com/1/941506444463212/project/1209173861012103/task/1213804797926971
Diff shows changes in other lines like *ETSEC, *MEVAL, and a lot of KID* variables. Could you please check.

jculhane:
> @jculhane A3 Q&V filename is missing the last "_" (in questions_and vars), so it is shown as a new file and hard to track the changes. Please re-upload it with the correct name and delete the incorrect file.
> 
Corrected.

> Also, want to double check that the updates are made only to the variables mentioned here? https://app.asana.com/1/941506444463212/project/1209173861012103/task/1213804797926971 Diff shows changes in other lines like *ETSEC, _MEVAL, and a lot of KID_ variables. Could you please check.

Yes - I'm not sure why it's showing all of those other variables as changes. The only changes are for the variables mentioned in the Asana task, to increase the max from 110 to 120.

hewanadun:
@jculhane Just noticed that A3 I4 plausibility checks need to be updated to do the comparison only if AGO and AGD variables are within the valid range (similar to IVP and FVP checks). I'm not sure why this wasn't caught before. Rule definitions are correct, could you please update the error checks.

hewanadun:
@echeng06 Please review the changes to derived variables. Thanks!

echeng06 (dismissed):
Derived changes look good to me!

@jculhane Related question though, there are a few age variables only in V3 in earlier that are still enforced to a maximum of 110. For the freeze logic, do we want to lift everything to 120 for consistency or keep it as-is? e.g.

* `cogflago`
* `bevhago`
* `beage`
* `parkage`
* `alsage`
* `moage`
* `decage` -> this one influences NACCCOGAGE so should change anyways

I don't have specific examples but I imagine there are cases for these variables that are over 110.

jculhane:
@echeng06 Let's keep the limit to 110 for v3 or prior variables.

jculhane:
> @jculhane Just noticed that A3 I4 plausibility checks need to be updated to do the comparison only if AGO and AGD variables are within the valid range (similar to IVP and FVP checks). I'm not sure why this wasn't caught before. Rule definitions are correct, could you please update the error checks.

@hewanadun updated!

hewanadun (approved):
Looks good! Thanks!

PR #537, opened 2026-06-03
Edits to A3 FVP plausibility checks

kgauthreaux:
Need to update these MOM/DAD ETPR checks to allow for 00 and 99 because if ETPR is equal to either, then the clinician leaves the rest of the row blank.(https://app.asana.com/1/941506444463212/project/1209173861012103/task/1215141809990490?focus=true)


hewanadun (approved):
Looks good!

PR #544, opened 2026-07-07
Edit form D1a alerts to errors for cognitive status - check at least one must be present.

kgauthreaux:
https://app.asana.com/1/941506444463212/project/1209173861012103/task/1215558688188410?focus=true 

Change the following QC checks from 'Alert' to 'Error'

d1a-ivp-p-1012
d1a-i4vp-p-1013
d1a-fvp-p-1029

EDIT: Actually also added the changes requested for this entry because I think to follow the TT rules I had to put in under D1a instead of D1b: https://app.asana.com/1/941506444463212/project/1209173861012103/task/1215558688188408?focus=true


hewanadun (approved):
Looks good! @kgauthreaux I fixed some missing commas as the CSV was not rendering properly. Please update the RT local copy. Thanks!

kgauthreaux:
> Looks good! @kgauthreaux I fixed some missing commas as the CSV was not rendering properly. Please update the RT local copy. Thanks!

Okay! We just pull the version down from Github and use that to edit. I think it's just Excel that kills us with adding extra things, but it only does it sometimes so hopefully it will be okay!

PR #549, opened 2026-08-06
Upload COVID F2 variable list and m/c checks

jculhane:
No plausibility checks


hewanadun (changes requested):
@jculhane Since we are going to merge F2 and F3 forms into one module, we need to make some changes to the header variables

- Move ADCID, PTID, VISITDATE, FORMVER, PACKET to header variables
- Add following variables at form level
   - FRMDATEF2, FRMDATEF3
   - INITIALSF2, INITIALSF2
   - MODEF2, MODEF3
   - F2NOT, F3NOT

Please revise the [F3 error checks](https://github.com/naccdata/uniform-data-set/pull/550) too.

hewanadun:
@jculhane Also please rename the FVP specific variables (like C19INF1BMO_F), since IVP and FVP forms are going to be displayed using branching logic in REDCap (no separate instruments), don't need a separate FVP variable name for common questions.

jculhane:
@hewanadun hopefully this fixes everything! I also uploaded the header CSVs here - will upload F3 in the separate pull request. I will be mostly OOO this week, so @kgauthreaux may be a faster source for quick edits.

hewanadun:
- Please add data_type column to Q&V CSVs
- Conformity column in header Q&V does not match how it is defined for other modules (i.e. for PACKET and FORMVER actual values are listed in other modules). FORMVER is numeric (not integer)
- Change all error codes to lowercase
- comp_vars column is not filled
- In FVP Q&V, C19RECDIAG should be a required variable?
- See the attached CSV for the issues found by the rule definition convertor. It only reports one place that issue was found but please fix all the places where applicable (i.e  variable name, test logic, sort desc, full desc, etc.)
[conversion_report_f2.csv](https://github.com/user-attachments/files/31486079/conversion_report_f2.csv)
- ~~C19TBST1, options doesn't match with REDCap for both IVP and FVP. Update Q&Vs and error checks~~ (REDCap form needs to be updated)
- C19INF1BYR_F, Q&V has min value 2019, conformity check has 2020

hewanadun:
@jculhane more changes
- [conversion_report_f2_090226.csv](https://github.com/user-attachments/files/31742311/conversion_report_f2_090226.csv)
- It seems the packet codes also changed to lowercase in error checks CSVs (while changing the error codes), please change the packet codes back to uppercase

hewanadun:
These variables are missing the conformity check in both CV and FCV checks, please add.
    C19VAC1ARM
    C19VAC1FLU
    C19VAC1MEM
    C19VAC1DEC
    C19VAC1UNK
    C19VAC2ARM
    C19VAC2FLU
    C19VAC2MEM
    C19VAC2DEC
    C19VAC2UNK
    C19BST1ARM
    C19BST1FLU
    C19BST1MEM
    C19BST1DEC
    C19BST1UNK

hewanadun (approved):
Looks good!

PR #550, opened 2026-08-06
Upload COVID F3 variable lists and m/c checks

jculhane:
No plausibility checks


hewanadun:
@jculhane A few more changes are required,
- Please add data_type column to Q&V CSVs
- Change all error codes to lowercase
- comp_vars column is not filled
- Change where it says "since [last f3_form_date]" in descriptions to "since the last visit"

hewanadun (approved):
Looks good!

